Issue #9, opened 2018-11-07
Introduce a Bootstrap theme


pgesek:
Select an existing, free Bootstrap theme and introduce it to the site in order to change the default bland JHipster look.

pgesek:
@oskarhinc  can you replace the orange on the home/login page with something more neutral?

oskarhinc:
Sure, how about blue?
![screenshot from 2018-11-13 07-54-34](https://user-images.githubusercontent.com/19305104/48396049-7650b700-e719-11e8-89bc-6f4b909acbf2.png)

pgesek:
I think it looks fine for now

pgesek:
Moving to done

PR #14, opened 2018-11-08
#9: Introduced custom Bootstrap theme

oskarhinc:


src/main/webapp/i18n/en/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "global": {
-    "title": "ServiceNet",
+    "title": " ServiceNet",
```
pgesek: Was this intentional? If yes, the space does not belong here

oskarhinc: It was intentional, to add a space after the logo in the header in the quickest way. Although, I can handle it with CSS.

Issue #10, opened 2018-11-07
Add MongoDb to the project


pgesek:
* Add MongoDB to docker configuration (compose)
* Add MongoDB connection configuration to config files
* Include MongoDB on the health checks page
* Introduce Spring Data code that can be later on used to consume Mongo: https://spring.io/projects/spring-data-mongodb

pgesek:
The code looks fine, we will test once we start storing docs in the db.

PR #15, opened 2018-11-13
#10: Added mongoDB to the project

oskarhinc:


src/main/java/org/benetech/servicenet/config/NoSQLDatabaseConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private static final int TIMEOUT_IN_MS = 500;
+
+    @Autowired
+    private Environment env;
+
+    @Bean
+    public MongoDbFactory mongoDbFactory() {
+        MongoClientOptions.Builder optionsBuilder = MongoClientOptions.builder();
+        optionsBuilder.serverSelectionTimeout(TIMEOUT_IN_MS);
```
pgesek: timeout should also come from props

pgesek: You should also be able to just inject a conf value:

@Value("${spring.data.mongodb.uri}")
private String mongoUrl;

oskarhinc: Right, I'll change the approach

Issue #7, opened 2018-11-06
Update the application to Java 11


oskarhinc:
JHipster doesn't support creating an app with Java 11, so it has to be converted manually. This link should be helpful: https://gist.github.com/mraible/716f4d2ab205c0e747d51119fc5d9cd8#gistcomment-2598658

pgesek:
@oskarhinc I verified this works, can we move to done?

oskarhinc:
@pgesek sure, I'll move it

PR #27, opened 2018-11-15
#7: Updated the application to Java 11

oskarhinc:
To make the app working on Java 11 I had to:
- remove SpotBugsPlugin, as it's not supported over Java 9
- change some of the dependencied versions
- add some dependencies to maven-compiler-plugin annotation processor paths
- override javassist dependency version provided by jhipster


pgesek:
I verified it works on Java 11

Issue #30, opened 2018-11-15
Capture updates to entities as Metadata


pgesek:
HSDS uses the metadata entity to capture changes to entities. Add a hibernate based component that processes db transactions and is able to register these metadata changes to any resource (entity) changing.

* Define Hibernate based interceptor/event handler/other to create Metadata entries for any change happening to data
* Resource id should have the UUID of the resource referenced

kcissewski:
Tested with success.

PR #34, opened 2018-11-26
#30: Added Hibernate Post Events Listeners

oskarhinc:


oskarhinc:
The integration tests need to be adjusted to cover those changes before the merge.

oskarhinc:
I've changed the mocking approach in the tests, and everything should work now.

src/test/java/org/benetech/servicenet/web/rest/ServiceTaxonomyResourceIntTest.java
```diff
@@ -48,6 +49,9 @@
     private static final String DEFAULT_TAXONOMY_DETAILS = "AAAAAAAAAA";
     private static final String UPDATED_TAXONOMY_DETAILS = "BBBBBBBBBB";
 
+    @Autowired
+    private HibernateInterceptor hibernateInterceptor;
```
pgesek: why do we have to disable it before every test?

oskarhinc: Because all metadata is connected to the currently logged user, and it would throw an error if no user is signed in.

oskarhinc: I was also thinking about authorizing user before each test, but I hadn't found a solution that would work.

pgesek: Implement a test user service that extends the regular user service, but overrides the method that you are using to get the user. Then add an additional configuration bean to your test classes. In that configuration redefine the userService bean with a you new test implementation.

Issue #35, opened 2018-11-26
Add backend for uploading files


oskarhinc:
All uploaded files should be sent to the server and be persisted in the non-relational database. Use the provided upload page for this. Only authorized users should be able to upload a file. If the file is in a different format than JSON, it should be converted before saving.

wojtekbu:
I've tested most of it here https://github.com/benetech/ServiceNet/issues/3
and also I have checked:
* persistence in MongoDB,
* content can be inspected on webapp.

PR #36, opened 2018-11-27
#35: Added backend fo file upload

oskarhinc:


src/main/java/org/benetech/servicenet/service/impl/DocumentUploadServiceImpl.java
```diff
@@ -25,22 +35,34 @@
...
+
+    @Override
+    public DocumentUploadDTO uploadFile(MultipartFile file, String delimiter) throws IllegalArgumentException, IOException {
+
+        String parsedDocumentId = null;
+        try {
+            AbstractFileConverter converter = FileConverterFactory.getConverter(file, delimiter);
+            parsedDocumentId = mongoDbService.saveParsedDocument(converter.convert(file));
+        } catch (IOException ex) {
+            log.error(ex.getMessage(), ex);
```
pgesek: I think we want to fail at this point, not just log the error?

oskarhinc: You're right, I'll change it.

Issue #38, opened 2018-11-27
Map different data files to a common structure


oskarhinc:
As known, data files from different providers have a different structure. We should map those structures to our entities and persist them in the database.

PR #39, opened 2018-12-04
#38: Added architecture for mapping data files to entities

oskarhinc:


src/main/java/org/benetech/servicenet/adapter/AbstractDataAdapter.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * @param data list of multiple objects of different types to be mapped and persisted, all in JSON format
+     */
+    public abstract void persistData(List<String> data);
+
+    /**
+     * @return true if only single objects mapping is supported
+     */
+    public abstract boolean isUsedForSingleObjects();
```
pgesek: rename to supportsMultipleObjects

src/main/java/org/benetech/servicenet/adapter/AbstractDataAdapter.java
```diff
@@ -0,0 +1,25 @@
...
+
+    protected static final String SINGLE_MAPPING_NOT_SUPPORTED = "Single objects mapping is not supported";
```
pgesek: can these messages be more clear? what is single/multiple mapping?

src/main/java/org/benetech/servicenet/adapter/shared/util/LocationUtils.java
```diff
@@ -0,0 +1,28 @@
...
+import org.benetech.servicenet.adapter.shared.model.Coordinates;
+
+import javax.validation.constraints.NotNull;
+import java.util.Optional;
+
+public final class LocationUtils {
+
+    public static Optional<Coordinates> getCoordinatesFromString(String data, String delimiter) {
+        try {
+            Coordinates result = new Coordinates();
```
pgesek: This parsing needs better null handling I think

src/main/java/org/benetech/servicenet/adapter/shared/util/LocationUtils.java
```diff
@@ -0,0 +1,28 @@
...
+            String[] parts = data.split(delimiter);
+            result.setLatitude(Double.valueOf(parts[0]));
+            result.setLongitude(Double.valueOf(parts[1]));
+            return Optional.of(result);
+        } catch (NumberFormatException | IndexOutOfBoundsException ex) {
```
pgesek: We don't really want to swallow exception like this, we should rather use conditional constructs to make sure parsing is correct

src/main/java/org/benetech/servicenet/adapter/firstprovider/FirstProviderDataMapper.java
```diff
@@ -0,0 +1,42 @@
...
+            c -> result.latitude(c.getLatitude()).longitude(c.getLongitude()))
+            .orElse(result);
+    }
+
+    default Phone extractPhone(RawData rawData) {
+        return new Phone().number(rawData.getPhone());
+    }
+
+    private String extractLocationName(RawData rawData) {
+        return LocationUtils.buildLocationName(rawData.getCity(), rawData.getStateProvince(), rawData.getAddress1());
```
pgesek: I think we need to give some more though on what we want as the location name, but you can leave this as is for now

src/test/java/org/benetech/servicenet/adapter/AdapterTestsUtils.java
```diff
@@ -0,0 +1,18 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+
+public final class AdapterTestsUtils {
+
+    private static final String RESOURCE_PACKAGE = "adapters/";
+
+    public static String readResourceAsString(String resourceName) throws IOException {
+        ClassLoader classloader = Thread.currentThread().getContextClassLoader();
+        InputStream is = classloader.getResourceAsStream(RESOURCE_PACKAGE + resourceName);
```
pgesek: Use try with resource  - always close the stream

Issue #40, opened 2018-12-07
Prepare matching algorithm architecture


oskarhinc:
The purpose of this issue is to prepare architecture for the matching algorithm. The usage of comparing methods should be used in tests for now. It will be plugged to the app later. See https://github.com/benetech/ServiceNet/wiki/Matching-algorithm-approaches for more details about the algorithm.

PR #41, opened 2018-12-10
#40: Added architecture for matching algotihm

oskarhinc:


src/main/java/org/benetech/servicenet/matching/AlternateNameSimilarityCounter.java
```diff
@@ -0,0 +1,18 @@
+package org.benetech.servicenet.matching;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AlternateNameSimilarityCounter extends AbstractSimilarityCounter<String> {
+
+    @Value("${similarity-ratio.weight.alternate-name}")
```
pgesek: I think it would be good to make the ratio independent from the similarity matcher - applying ratios should be the responsibility of the aggregator.

Issue #31, opened 2018-11-16
Create an admin usertype that can upload files as a different user/org


johnhbenetech:

wojtekbu:
Tested on backend and frontend:
* a USER can upload files only for an associated system account,
* an ADMIN can upload files for all associated system accounts,
* an ANONYMOUS USER cannot upload any files.
Closing

PR #50, opened 2018-12-17
#31: Provider name is now dependent of user role

oskarhinc:


src/main/java/org/benetech/servicenet/adapter/DataAdapterFactory.java
```diff
@@ -18,15 +19,15 @@ public DataAdapterFactory(ApplicationContext context) {
     public Optional<SingleDataAdapter> getSingleDataAdapter(String providerName) {
         try {
             return Optional.of(context.getBean(providerName + DATA_ADAPTER_SUFFIX, SingleDataAdapter.class));
-        } catch (NoSuchBeanDefinitionException e) {
+        } catch (NoSuchBeanDefinitionException | BeanNotOfRequiredTypeException e) {
```
pgesek: Is this new exception something we handle silently like this? Seems like a configuration issue we would need to signal on

src/main/java/org/benetech/servicenet/adapter/DataAdapterFactory.java
```diff
@@ -18,15 +19,15 @@ public DataAdapterFactory(ApplicationContext context) {
...
             return Optional.empty();
         }
     }
 
     public Optional<MultipleDataAdapter> getMultipleDataAdapter(String providerName) {
         try {
             return Optional.of(context.getBean(providerName + DATA_ADAPTER_SUFFIX, MultipleDataAdapter.class));
-        } catch (NoSuchBeanDefinitionException e) {
+        } catch (NoSuchBeanDefinitionException | BeanNotOfRequiredTypeException e) {
```
pgesek: same as above

src/main/java/org/benetech/servicenet/service/UserService.java
```diff
@@ -318,6 +318,15 @@ public void removeNotActivatedUsers() {
         return authorityRepository.findAll().stream().map(Authority::getName).collect(Collectors.toList());
     }
 
+    /**
+     * @return true if any user is logged, and he has admin role
+     */
+    public boolean isAdmin() {
+        return getUserWithAuthorities().map(u -> u.getAuthorities().stream()
```
pgesek: isAdmin can be a method of the user or UserDTO entity

src/main/java/org/benetech/servicenet/service/UserService.java
```diff
@@ -318,6 +318,15 @@ public void removeNotActivatedUsers() {
         return authorityRepository.findAll().stream().map(Authority::getName).collect(Collectors.toList());
     }
 
+    /**
+     * @return true if any user is logged, and he has admin role
+     */
+    public boolean isAdmin() {
```
pgesek: we should rename this method to: isCurrentUserAdmin

src/main/java/org/benetech/servicenet/service/impl/DocumentUploadServiceImpl.java
```diff
@@ -114,4 +114,15 @@ public void delete(UUID id) {
...
         documentUploadRepository.deleteById(id);
     }
+
+    private String getRealProviderName(String currentProviderName) {
+        if (userService.isAdmin()) {
+            return currentProviderName;
+        }
+
+        //TODO: Replace with provider name connected to the user
+        return userService.getUserWithAuthorities().map(User::getLogin).orElseThrow(
```
pgesek: this piece of code is strange, but I guess it's a todo?

src/main/webapp/app/shared/auth/private-route.tsx
```diff
@@ -7,6 +7,7 @@ import ErrorBoundary from 'app/shared/error/error-boundary';
 
 interface IOwnProps extends RouteProps {
   hasAnyAuthorities?: string[];
+  isAdmin: any;
```
pgesek: isn't this a boolean?

src/main/java/org/benetech/servicenet/adapter/DataAdapterFactory.java
oskarhinc: BeanNotOfRequiredTypeException will be thrown when we'll ask for SingleDataAdapter, but the adapter for this partner is a type of MultipleData. In this case, we're simply returning an empty optional. Should it be changed?

src/main/java/org/benetech/servicenet/service/impl/DocumentUploadServiceImpl.java
oskarhinc: Yes, when User will be connected to the System Account, we will return its name here, but for now, it's just login

src/main/webapp/app/shared/auth/private-route.tsx
oskarhinc: it is, thanks

src/main/java/org/benetech/servicenet/service/UserService.java
oskarhinc: ok

src/main/java/org/benetech/servicenet/service/UserService.java
oskarhinc: ok

src/main/java/org/benetech/servicenet/adapter/DataAdapterFactory.java
pgesek: If this is something that should not happen in normal usage (and I assume it shouldn't?) we should raise an exception

oskarhinc: Ok, thanks

src/main/java/org/benetech/servicenet/service/impl/DocumentUploadServiceImpl.java
oskarhinc: As it's already done, I've updated this PR to cover this.

src/main/java/org/benetech/servicenet/adapter/DataAdapterFactory.java
```diff
@@ -17,18 +16,28 @@ public DataAdapterFactory(ApplicationContext context) {
     }
 
     public Optional<SingleDataAdapter> getSingleDataAdapter(String providerName) {
-        try {
-            return Optional.of(context.getBean(providerName + DATA_ADAPTER_SUFFIX, SingleDataAdapter.class));
-        } catch (NoSuchBeanDefinitionException | BeanNotOfRequiredTypeException e) {
+        AbstractDataAdapter adapter = getDataAdapter(providerName);
+        if (adapter instanceof SingleDataAdapter) {
+            return Optional.of((SingleDataAdapter) adapter);
```
pgesek: You should use optional.ofNullable

oskarhinc: I think it's not necessary, as it's already verified and can't be null (by instanceof).

pgesek: Ok

Issue #51, opened 2018-12-17
Implement names matching algorithm


oskarhinc:
See https://github.com/benetech/ServiceNet/wiki/Matching-algorithm-approaches for details, and implement name and alternate name matching algorithms.

PR #62, opened 2018-12-19
#51: Added names matching algorithm

oskarhinc:


src/main/java/org/benetech/servicenet/matching/StringMatchingUtils.java
```diff
@@ -0,0 +1,25 @@
...
+        return name.toUpperCase(Locale.ROOT).replaceAll("[^A-Z0-9 ]", "").trim().replaceAll(" +", " ");
+    }
+
+    public static String sort(String name) {
+        String delimiter = " ";
+        return Arrays.stream(name.split(delimiter)).sorted().collect(Collectors.joining(delimiter));
+    }
+
+    public static String extractInitials(String name) {
+        String delimiter = " ";
```
wojtekbu: the same delimiter twice. please make it final

oskarhinc: Wow, nice catch!

Issue #43, opened 2018-12-10
Add a list of activities to the main page


wojtekbu:
The parent ticket can be found here: https://github.com/benetech/ServiceNet/issues/11
According to the mock https://user-images.githubusercontent.com/12927553/48443971-55f41d00-e758-11e8-96ad-6596c7a6092c.png add a list of all resources that have been uploaded/read from the logged-in Database Manager’s system.

Acceptance criteria:
- the list should show the same level of information as on the mock,
- the list does not have to look exactly the same as the mock,

wojtekbu:
@johnhbenetech I have a question regarding the mock https://user-images.githubusercontent.com/12927553/48443971-55f41d00-e758-11e8-96ad-6596c7a6092c.png
Should an 'x' button reject the updates or just postpone it?
Should we display all resources on the activities' list or just the updated ones?

johnhbenetech:
I think we should display all resources but always sort those with updates to the top of the list.

As for the 'x', I'm open to suggestions. I think the ideal interaction would be that clicking the 'x' hides/deprioritizes that record until a new conflict/update is found. this would probably require a new meta field like UserID/RecordID/NoUpdateBeforeDate or something. Seems like a lot of hassle for right now.

I think you can just break the 'x' into a different task and we can revisit it after the main features are built.

wojtekbu:
@johnhbenetech  according to the update meeting 12/11/18:

- every notification element on the list will have a full list of activities. We will revisit this solution later adjust it for example by adding unfold button,
- the list should be available on one page with a possibility to load more notifications. An amount of notification per one fetch will be adjusted in next iterations,
- blue - there is a conflict for resources which we believe are a match (synchronization for the first time),
- yellow/orange - there is a conflict for resources which were synchronized at least once,
- red - the same situation as for 'yellow/orange' but more than one provider decided to approve this version.

wojtekbu:
Services types are not available yet so tag like 'Housing' will be added later

oskarhinc:
I can confirm it's showing required information for mocked data

PR #82, opened 2019-01-04
#43:Added list of activities to the main page

wojtekbu:


src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -47,13 +48,14 @@ public ActivityServiceImpl(OrganizationService organizationService, ConflictServ
     @Override
     public List<ActivityDTO> getAllOrganizationActivities(UUID systemAccountId) {
         List<ActivityDTO> activities = new ArrayList<>();
-        List<OrganizationDTO> orgs = organizationService.findAllWithOwnerId(systemAccountId);
+        //List<OrganizationDTO> orgs = organizationService.findAllWithOwnerId(systemAccountId);
```
oskarhinc: Is it intentional?

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -47,13 +48,14 @@ public ActivityServiceImpl(OrganizationService organizationService, ConflictServ
...
+        List<OrganizationDTO> orgs = organizationService.findAll();
 
         // TODO: get organization-entities mapping for every organization
         // this mock below will be removed when Organization matching will be available
         List<UUID> entities = Collections.singletonList(UUID.fromString(EXAMPLE_DATA));
         UUID orgId = null;
-        if (orgs.get(0) != null) {
+        if (!CollectionUtils.isEmpty(orgs) && orgs.get(0) != null) {
```
oskarhinc: I think orgs.get(0) will never be null

src/main/webapp/app/modules/home/ActivityElement.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { APP_DATE_FORMAT } from 'app/config/constants';
+
+const ActivityElement = props => (
+  <Row style={{ marginBottom: '15px' }}>
+    <Col>
+      <CardGroup>
+        <Card style={{ marginBottom: '0px' }}>
+          <CardBody>
+            <CardTitle style={{ color: '#66c4ff' }}>
```
oskarhinc: We should avoid inline styles, especially with color values. Please use scss with variables here

src/main/webapp/app/modules/home/ActivityElement.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <CardTitle>
+              <FontAwesomeIcon icon="map-marker-alt" /> {props.activity.organization.locationName}
+            </CardTitle>
+          </CardBody>
+        </Card>
+        <Card style={{ backgroundColor: '#f4f4f4' }}>
+          <CardBody>
+            {props.activity.conflicts.map(conflict => (
+              <CardTitle style={{ color: '#66c4ff' }}>
+                {conflict.ownerName} {conflict.acceptedThisChanges.length > 0 && <b>+{conflict.acceptedThisChanges.length}</b>}
```
oskarhinc: you could put this value to the var, to be more readable

src/main/webapp/app/modules/home/ActivityElement.tsx
```diff
@@ -0,0 +1,59 @@
...
+                {conflict.acceptedThisChanges.length > 1 && (
+                  <Translate
+                    contentKey="serviceNetApp.activity.unresolved.conflictsPlusMany"
+                    interpolate={{ fieldName: conflict.fieldName }}
+                  >
+                    partners have a conflicting {conflict.fieldName}
+                  </Translate>
+                )}
+                {conflict.acceptedThisChanges.length === 1 && (
```
oskarhinc: You could put those components as a variables above, as it would be more readable:

`let conflictsPlusMany =  <Translate
                    contentKey="serviceNetApp.activity.unresolved.conflictsPlusMany"
                    interpolate={{ fieldName: conflict.fieldName }}
                  >
                    partners have a conflicting {conflict.fieldName}
                  </Translate>`
and then something like
`{conflict.acceptedThisChanges.length > 1 && conflictsPlusMany`

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
wojtekbu: It's a part of tmp mock so I'll leave it just in case

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
wojtekbu: Yes, I'll add //TODO

src/main/webapp/app/modules/home/ActivityElement.tsx
wojtekbu: ok

src/main/webapp/app/modules/home/ActivityElement.tsx
wojtekbu: I cannot because it's a list's variable.

src/main/webapp/app/modules/home/ActivityElement.tsx
wojtekbu: I cannot because it's a list's variable.

Issue #87, opened 2019-01-07
Connect the Organization matching algorithm with data upload


oskarhinc:
During/After data upload, the organization matching algorithm should be triggered. It should be executed in a separate thread. Each Newly created/updated organization should be compared with other ones, and OrganizationMatch entities should be created as a result. For each pair of organizations, there should be only one pair of entities.

wojtekbu:
Works fine

PR #91, opened 2019-01-07
 #87: Connected OrganizationMatching generation with data import

oskarhinc:


src/main/java/org/benetech/servicenet/domain/OrganizationMatch.java
```diff
@@ -184,11 +101,8 @@ public int hashCode() {
     public String toString() {
```
wojtekbu: please remove toString since @ Data = @ getter, @ Setter, @ tostring, @ EqualsAndHashCode and @ RequiredArgsConstructor

src/main/java/org/benetech/servicenet/service/dto/OrganizationMatchDTO.java
```diff
@@ -137,11 +56,8 @@ public int hashCode() {
     public String toString() {
```
wojtekbu: please remove toString since @ Data = @ getter, @ Setter, @ tostring, @ EqualsAndHashCode and @ RequiredArgsConstructor

wojtekbu:
@oskarhinc please, resolve the conflict

Issue #63, opened 2018-12-19
Load partner names dynamically on file upload dropdown


oskarhinc:
The list of partners to select should be loaded dynamically from the backend (all system accounts). As a part of this ticket, all partners from stage 1 of the project should be loaded to the database automatically when the app starts.

wojtekbu:
The functionality works but the list of all system accounts should be available only for admin. At this moment a default user can fetch, add, update and delete a system account. The 'USER' should be allowed only to get his/her system account.

wojtekbu:
@oskarhinc please add _@PreAuthorize("hasRole(\"" + AuthoritiesConstants.ADMIN + "\")")_ to
https://github.com/benetech/ServiceNet/blob/master/src/main/java/org/benetech/servicenet/web/rest/SystemAccountResource.java

wojtekbu:
@oskarhinc  Default user has no longer access to the list of accounts but he is still requesting it while entering the upload page.

wojtekbu:
Works fine

PR #93, opened 2019-01-08
 #63: Secured SystemAccount resource

oskarhinc:


src/main/java/org/benetech/servicenet/web/rest/SystemAccountResource.java
```diff
@@ -30,6 +32,7 @@
  */
 @RestController
 @RequestMapping("/api")
+@PreAuthorize("hasRole(\"" + AuthoritiesConstants.ADMIN + "\")")
```
wojtekbu: Will the USER be able to get info about its own system account?

oskarhinc: not via REST api

Issue #89, opened 2019-01-07
Add mapping services' entities between organizations


wojtekbu:
In order to detect conflicts, we have to rely on mappings between entities. For example, an organization can have multiple services and its mirror organization (from OrganizationMatch) also can have multiple services, so we have to figure it out which service from one organization represents a service from the other.
The solution might be a map with identifiers of entities and a full class name. This map will we used to detect conflicts.

ToDo:
* add a service which will create the map for  services' entities,
* use the service asynchronically right after the OrganizationMatch is found,
* handle detecting an entity match similarly to OrganizationMatch.

### Mappings:

Organization:
- Location location **→see below**,
- Organization replacedBy,
- DocumentUpload sourceDocument,
- SystemAccount account,
- Funding funding,
- **Set**\<Service\> services **→see below**.

Location:
- PhysicalAddress physicalAddress,
- PostalAddress postalAddress,
- RegularSchedule regularSchedule **→see below**,
- HolidaySchedule holidaySchedule,
- **Set**\<Language\> langs,
- **Set**\<AccessibilityForDisabilities\> accessibilities.

Service:
- ServiceAtLocation location **→see below**,
- RegularSchedule regularSchedule **→see below**,
- HolidaySchedule holidaySchedule,
- Funding funding,
- Eligibility eligibility,
- **Set**\<ServiceArea\> areas,
- **Set**\<RequiredDocument\> docs,
- **Set**\<PaymentAccepted\> paymentsAccepteds,
- **Set**\<Language> langs,
- **Set**\<ServiceTaxonomy\> taxonomies,
- **Set**\<Phone\> phones.

ServiceAtLocation:
- RegularSchedule regularSchedule,
- HolidaySchedule holidaySchedule,
- Set\<Phone\> phones.

RegularSchedule:
- **Set**\<OpeningHours\> openingHours.

oskarhinc:
@wojtekbu for now, only Organization fields should be mapped

PR #96, opened 2019-01-09
#89: Added architecture and tests for finding entities equivalents

oskarhinc:


src/main/java/org/benetech/servicenet/matching/model/EntityEquivalent.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.Set;
+import java.util.UUID;
+
+@Data
+@NoArgsConstructor
+@RequiredArgsConstructor
+public class EntityEquivalent {
+
+    @SuppressWarnings("checkstyle:visibilitymodifier")
+    protected Set<EntityEquivalent> otherEquivalents = new HashSet<>();
```
wojtekbu: @ Builder.Default

Issue #102, opened 2019-01-09
Add 'compare' section to the single record view


oskarhinc:
See https://user-images.githubusercontent.com/12927553/48446723-9b681880-e75f-11e8-867b-4ffe6760ca4c.png
The section should contain all records that seem to be matched organizations. The list of matched organizations to compare could be visible in the section, as it shouldn't be too long. Each element should be a link to the side-by-side view (which should be done in a separate issue: https://github.com/benetech/ServiceNet/issues/113)

kcissewski:
Tested with success.

PR #117, opened 2019-01-15
 #102: Added 'compare' section to the single record view

oskarhinc:


src/main/webapp/i18n/en/singleRecordView.json
```diff
@@ -5,7 +5,14 @@
       "otherTabs": "Other tabs..."
     },
     "details": {
-      "title": "Organization details"
+      "title": "Organization details",
+      "compare": {
+        "title": "Compare",
+        "header": " partners have similar records",
+        "content1": "Reviewing these records will help us give you better recommendation in the future.",
+        "content2": "A manual review is the easiest way to confirm that these records match",
```
kcissewski: I think there should be a period at the end of a sentence.

oskarhinc: true

Issue #56, opened 2018-12-17
Map data from ShelterTech to a common structure


oskarhinc:
An adapter for data from partner ShelterTech should be implemented, with a wiki page, describing how it's mapped.

oskarhinc:
I confirm the data is retrieved and mapped

PR #146, opened 2019-01-22
#56 Added mapping for ShelterTech

wojtekbu:


src/main/java/org/benetech/servicenet/adapter/sheltertech/mapper/ShelterTechOrganizationMapper.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Named("emailFromString")
+    default String emailFromString(String emailString) {
+        Pattern email = Pattern.compile("[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,6}", Pattern.CASE_INSENSITIVE);
+
+        if (StringUtils.isNotBlank(emailString)) {
+            Matcher matcher = email.matcher(emailString);
+
+            // Consider supporting many e-mail addresses
```
oskarhinc: add TODO as it will be easier to get back to

src/main/java/org/benetech/servicenet/adapter/sheltertech/mapper/ShelterTechOrganizationMapper.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Named("locationFromAddressRaw")
+    default Location locationFromAddressRaw(AddressRaw raw) {
+        return ShelterTechLocationMapper.INSTANCE.mapToLocation(raw);
+    }
+
+    @Named("emailFromString")
+    default String emailFromString(String emailString) {
+        Pattern email = Pattern.compile("[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,6}", Pattern.CASE_INSENSITIVE);
```
oskarhinc: I think a little comment how this pattern works would be great here

src/main/java/org/benetech/servicenet/adapter/sheltertech/mapper/ShelterTechPhoneMapper.java
```diff
@@ -0,0 +1,38 @@
...
+    @Mapping(ignore = true, target = "description")
+    @Mapping(ignore = true, target = "contact")
+    Phone mapToPhone(PhoneRaw raw);
+
+    List<Phone> mapToPhones(List<PhoneRaw> phonesRaw);
+
+    @Named("languageFromCountryCode")
+    default String languageFromCountryCode(String countryCode) {
+        if (StringUtils.isNotBlank(countryCode)) {
+            if (StringUtils.equalsIgnoreCase("US", countryCode)) {
```
oskarhinc: if it will be null, equals will return false so there's no need to verify it before

src/main/java/org/benetech/servicenet/adapter/sheltertech/mapper/ShelterTechRegularScheduleMapper.java
```diff
@@ -0,0 +1,96 @@
...
+
+    default String getFormattedTime(String nonFormatted) {
+        String result;
+        Pattern fullHourSchema = Pattern.compile("[0-9]{4}");
+        Matcher fullHourMatcher = fullHourSchema.matcher(nonFormatted);
+        Pattern hourSchema = Pattern.compile("[0-9]{3}");
+        Matcher hourMatcher = hourSchema.matcher(nonFormatted);
+
+        if (fullHourMatcher.find()) {
+            String hourAndMin = fullHourMatcher.group();
```
oskarhinc: I think those actions could be moved to a separate method to indicate what they are doing more easily - e. g. mapToTwoDigitHours() or sth like that

src/main/java/org/benetech/servicenet/domain/Phone.java
```diff
@@ -60,138 +65,55 @@
     @JsonIgnoreProperties("phones")
     private ServiceAtLocation serviceAtLocation;
 
-    // jhipster-needle-entity-add-field - JHipster will add fields here, do not remove
```
oskarhinc: Are you sure jhipster won't fail it the needle will be absent? Will it just omit generating getters and setters?

src/main/java/org/benetech/servicenet/domain/RequiredDocument.java
```diff
@@ -40,34 +46,15 @@
     @Column(name = "provider_name")
     private String providerName;
 
-    // jhipster-needle-entity-add-field - JHipster will add fields here, do not remove
```
oskarhinc: same here

src/main/java/org/benetech/servicenet/service/impl/ImportServiceImpl.java
```diff
@@ -190,22 +206,25 @@ public Service createOrUpdateService(Service service, String externalDbId, Strin
...
-        return phones;
+    @Override
+    @Transactional
+    public Set<Phone> createOrUpdatePhonesForOrganization(Set<Phone> phones, UUID orgId) {
+        Set<Phone> orgPhones = phoneRepository.findAllByOrganization(orgId);
+        return persistPhones(phones, orgPhones);
     }
 
     @Override
+    @Transactional
```
oskarhinc: I think we can annotate the whole class instead of separate methods

src/main/resources/config/application.yml
```diff
@@ -62,7 +62,7 @@ spring:
         mode: HTML
     servlet.multipart:
         max-request-size: 100MB
-        max-file-size: 5MB
+        max-file-size: 6MB
```
oskarhinc: As it's changed so often, maybe we should increase it to safe 10MB or so?

Issue #168, opened 2019-01-31
Add unit tests for LAAC adapter and mapping


oskarhinc:
As described in #135 we need to add unit tests to verify the server works well for both missing and complete data.

oskarhinc:
All issues found by the tests should be fixed in #177

PR #169, opened 2019-01-31
#168: Added test for complete data mapping

oskarhinc:


src/test/java/org/benetech/servicenet/adapter/laac/LAACCompleteDataAdapterTest.java
```diff
@@ -0,0 +1,173 @@
...
+        when(importService.createOrUpdateLocation(any(Location.class), anyString(), anyString()))
+            .thenReturn(new Location());
+
+        when(importService.createOrUpdateOrganization(any(Organization.class), anyString(), anyString(),
+            any(DataImportReport.class)))
+            .thenReturn(new Organization());
+
+        String json = AdapterTestsUtils.readResourceAsString(COMPLETE_JSON);
+        SingleImportData importData = new SingleImportData(json, new DataImportReport(), PROVIDER_NAME, true);
+        adapter.importData(importData);
```
kcissewski: This is not part of the setup process and should be included in each test method.

Issue #47, opened 2018-12-10
Add sorting activities to the main page


wojtekbu:
The parent ticket can be found here: https://github.com/benetech/ServiceNet/issues/11
According to the mock https://user-images.githubusercontent.com/12927553/48443971-55f41d00-e758-11e8-96ad-6596c7a6092c.png add sorting functionality.

Default sort should be based on the number and age of updates of matching records in partners’ databases. 

Sorting options:
-  the number of updates,
 - age of updates.

johnhbenetech:
also see #18 for more info

oskarhinc:
@wojtekbu I've verified it, and it's working well, but I've got some suggestions:

1. I think it should be available to determine if the sorting is ascending or descending, even as different dropdown options:
-by xyz (ascending)
-by xyz (descending)

2. The layout of the button could be quite different (it should not stick with the first activity on the bottom). Also for me, it's more intuitive for sort button to be floated right, but I'll leave it to you.    

3. I think it's a good solution to display the selected option all the time (like it's done for example on the amazon.com: 
![screenshot from 2019-02-08 13-56-06](https://user-images.githubusercontent.com/19305104/52479567-624a1380-2ba9-11e9-8178-4be1651c5c29.png)

johnhbenetech:
As discussed in our Call:

* Design should be more in line with mock-up. Keep dropdown on same level as title and float right

* Sort Options should be:
**Recently Updated
**Recommended

* No need for ascending/descending option. Default should be newest to oldest, most to least.

Recommended sort at first will just be sorted by Number of Updated fields (most to least). In the future we will add weighting to fields updated, and weighting to updates over existing conflicts.

oskarhinc:
The sorting looks as expected, but the Last updated date seems to be always `now()`. @wojtekbu do you know why it might be happening? If so, could you create a ticket for that?

wojtekbu:
@oskarhinc It looks like conflicts cannot find a date for conflicted objects. Here is a solution https://github.com/benetech/ServiceNet/issues/239

PR #194, opened 2019-02-06
#47: Added sorting to the main page

wojtekbu:


src/main/java/org/benetech/servicenet/service/comparator/ActivityComparatorFactory.java
```diff
@@ -0,0 +1,24 @@
...
+public final class ActivityComparatorFactory {
+
+    private static final String SIZE = "size";
+
+    private static final String AGE = "age";
+
+    public static Comparator<ActivityDTO> createComparator(Pageable pageable) {
+        if (pageable.getSort().getOrderFor(AGE) != null) {
+            return new ActivityUpdateAgeComparator();
+        } else {
```
oskarhinc: either make size default, or add 'else if SIZE', and another default

src/main/webapp/app/modules/home/home.tsx
```diff
@@ -31,6 +41,18 @@ export class Home extends React.Component<IHomeProp, IHomeState> {
     }
   }
 
+  sort = prop => () => {
+    this.setState({ sort: prop }, () => {
+      this.reset();
+    });
+  };
+
+  toggleSort() {
```
oskarhinc: Use arrow function, so you don't have to bind this method

Issue #189, opened 2019-02-05
Update docker images 


oskarhinc:
For some reason, the docker images are not built properly now. Investigate the issue and fix it.

wojtekbu:
@oskarhinc please set a fixed MongoDB's version

wojtekbu:
Works fine

PR #210, opened 2019-02-08
#189: Specified mongodb image version

oskarhinc:


src/main/docker/mongo.yml
```diff
@@ -3,7 +3,7 @@ version: '2'
 services:
 
     servicenet-mongo:
-        image: mongo
+        image: mongo:3.4.19
```
wojtekbu: @oskarhinc why do we use the older version? The current stable version is 4.0 https://docs.mongodb.com/manual/release-notes/

oskarhinc: my mistake

Issue #175, opened 2019-01-31
Adjust logic of iCarolDataAdapter to new ImportService implementation


kcissewski:
This ticket has to be done after #170 

Since ImportService will change we need to adjust logic to the new implementation.

wojtekbu:
I have tested it on iCarol API data for Eden for period 2019/02/10-2019/02/18 and it works fine

wojtekbu:
https://github.com/benetech/ServiceNet/issues/176 main issue

PR #237, opened 2019-02-15
#175: Adjusted logic of ICarolAdataAdapted to new ImportService

oskarhinc:


src/main/java/org/benetech/servicenet/adapter/icarol/RelationManager.java
```diff
@@ -73,30 +68,58 @@ private static void logProgress(int itemNr, int size) {
...
 
-    private void saveServiceRelatedData(Location location, ICarolProgram program,
-                                        Service savedService) {
-        persistence.importEligibility(program, savedService);
+    private Optional<Location> getLocationToPersist(ImportData importData, ICarolSite site) {
+        return mapper.extractLocation(
+            site.getContactDetails(), site.getId(), importData.getProviderName())
+            .flatMap(extractedLocation -> {
+
+                extractedLocation.setPhysicalAddress(
```
wojtekbu: empty line

src/main/java/org/benetech/servicenet/adapter/icarol/RelationManager.java
```diff
@@ -73,30 +68,58 @@ private static void logProgress(int itemNr, int size) {
...
+                    mapper.extractPhysicalAddress(site.getContactDetails()).orElse(null));
+
+                extractedLocation.setPostalAddress(
```
wojtekbu: empty line

src/main/java/org/benetech/servicenet/adapter/icarol/RelationManager.java
```diff
@@ -73,30 +68,58 @@ private static void logProgress(int itemNr, int size) {
...
     }
 
-    private void saveEntitiesWithoutOrganization(ICarolDataToPersist data, ImportData importData) {
-        saveLocationsAndRelatedData(data.getSites(), data, importData, null, null);
-        saveServicesAndRelatedData(null, null, data.getPrograms(), importData);
+    private Optional<Service> getServiceToPersist(ImportData importData, ICarolProgram program) {
+        Service extractedService = mapper
+            .extractService(program, importData.getProviderName());
+
+        extractedService.setEligibility(mapper.extractEligibility(program).orElse(null));
```
wojtekbu: https://github.com/benetech/ServiceNet/wiki/iCarol-Data-Mapping-(Eden-and-UWBA)#restrictions

Eligibility cannot be null

oskarhinc: Indeed, that's why we set the whole Eligibility entity/relation to null. If mapper can't extract it, we shouldn't persist Empty eligibility.

src/main/java/org/benetech/servicenet/adapter/icarol/RelationManager.java
oskarhinc: ok

Issue #172, opened 2019-01-31
Adjust logic of HealthleadsDataAdapter to new ImportService implementation


kcissewski:
This ticket has to be done after #170 

Since ImportService will change we need to adjust logic to the new implementation.

wojtekbu:
I have tested it on the Healthleads real data from 2019-01-02 and it works fine

wojtekbu:
https://github.com/benetech/ServiceNet/issues/176 main issue

PR #242, opened 2019-02-18
#172: Adjusted Logic of healthleads to new ImportService

oskarhinc:


src/main/java/org/benetech/servicenet/adapter/healthleads/persistence/PersistanceManager.java
```diff
@@ -0,0 +1,198 @@
...
+    private Set<ServiceTaxonomy> getTaxonomiesToPersist(String serviceId) {
+        Set<HealthleadsServiceTaxonomy> serviceTaxonomies = dictionary.getRelatedEntities(
+            HealthleadsServiceTaxonomy.class, serviceId, HealthleadsService.class);
+
+        Set<ServiceTaxonomy> result = new HashSet<>();
+        for (HealthleadsServiceTaxonomy healthleadsServiceTaxonomy : serviceTaxonomies) {
+            ServiceTaxonomy serviceTaxonomy = mapper.extractServiceTaxonomy(healthleadsServiceTaxonomy)
+                .providerName(PROVIDER_NAME).externalDbId(serviceId);
+            Set<HealthleadsTaxonomy> taxonomies = dictionary.getRelatedEntities(
+                HealthleadsTaxonomy.class, healthleadsServiceTaxonomy.getTaxonomyId(), HealthleadsBaseData.class);
```
wojtekbu: Why did you use HealthleadsBaseData class instead of HealthleadsServiceTaxonomy class?

src/main/java/org/benetech/servicenet/adapter/healthleads/persistence/PersistanceManager.java
```diff
@@ -0,0 +1,198 @@
...
+        this.mapper = HealthLeadsDataMapper.INSTANCE;
+        this.importManager = importManager;
+    }
+
+    public void addData(HealthleadsBaseData data) {
+        if (isLocationBased(data)) {
+            dictionary.addData(data.getLocationId(), data, HealthleadsLocation.class);
+        } else if (isServiceBased(data)) {
+            dictionary.addData(data.getServiceId(), data, HealthleadsService.class);
+        } else if (isOrganizationBased(data)) {
```
wojtekbu: are you sure of using 'else if' instead of 'if'? Do isLocationBased, isLocationBased and isOrganizationBased cases are always disjointed?

oskarhinc: We're not sure of that, but each data we've seen was constructed this way.

src/main/java/org/benetech/servicenet/adapter/healthleads/persistence/PersistanceManager.java
oskarhinc: It should be done this way as all data that are not Location/Organization/Service related, are just considered to be HealthleadsBaseData related (line 54) as specifying it precisely would bring no benefits here.

Issue #187, opened 2019-02-05
Make the home page pageable


oskarhinc:
If there are many conflicts found for the provider, the home page loads too slow. There should be pagination for the activities to display.

UPDATE 02/15/19:
* a user should always see how many activities are displayed and what is the total number of them,
* pagination should be implemented as a button 'see xyz more' at the bottom https://user-images.githubusercontent.com/12927553/48443971-55f41d00-e758-11e8-96ad-6596c7a6092c.png
* add infinite scroll instead of button if possible

wojtekbu:
@oskarhinc Here is a PR for that feature https://github.com/benetech/ServiceNet/pull/245
Could you test it after review?

oskarhinc:
@wojtekbu the pagination can't be done on the front-end side as it doesn't speed loading the records at all. All records are loaded anyway and it took too long. The infinite scroll would look really cool but we should indicate the user that more records are loading, or it could be omitted if it would load really fast, but for now, I think it's confusing for the user that not all records were loaded, and he has no possibility to load more. The best performance we could achieve with using pagination on the DB but as we have to create the records on the fly I think it's fine to just do it within the algorithm manually.

oskarhinc:
I confirm it's working well, but I think it would be nice to indicate that the records are loading, as at the beginning for a few seconds the user only sees 0/0. Also maybe we could decrease the number of fetched records to 5 or so? However, I don't know if that's a priority now. 

@johnhbenetech could you verify if that satisfies you, or you would like us to change something?

PR #249, opened 2019-02-19
#187: Moved pagination logic to the database

wojtekbu:


src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -55,66 +54,44 @@ public ActivityServiceImpl(OrganizationService organizationService, ConflictServ
...
+        for (Organization org : orgs) {
             try {
-                Optional<ActivityDTO> activityOpt = getEntityActivity(org.getId(), org.getId());
+                Optional<ActivityDTO> activityOpt = getEntityActivity(org.getId());
                 activityOpt.ifPresent(activities::add);
             } catch (ActivityCreationException ex) {
                 log.error(ex.getMessage());
             }
         }
+        Comparator<ActivityDTO> comparator = ActivityComparatorFactory.createComparator(pageable);
```
oskarhinc: Sorting done after fetching batch from the db might result in inaccurately sorted records.

Issue #220, opened 2019-02-11
Ensure that entities like Opening Hours are not detached


oskarhinc:
It seems like in some cases entities like OpeningHours has relation no with Regular Schedule or HolidaySchedule either. We need to verify this and add a proper logic to the ImportService. We also need tests verifying if it works properly for the future and other entities ( like Phones, Languages, and other list based)

oskarhinc:
Additionally, we need to determine if we want to store historical data for some purposes, or we can get rid of them, as we have metadata anyway. Based on the decision, we need to add the same cascade removal for detached entities and test it well.

johnhbenetech:
Closing this as obsolete

PR #250, opened 2019-02-19
#174: Removed old import service

oskarhinc:


oskarhinc:
@wojtekbu, please pay extra attention to assertions ins tests, made in commit ` [66740a4](/benetech/ServiceNet/pull/250/commits/66740a41252f411c5da0c2925d00d68a35828fa1)`. Due to the change of the import service, I had to rewrite them. I've verified it multiple times, but please verify if you agree with them as it's important functionality.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -109,142 +172,113 @@ public void shouldNotCreateLocationIfConfidential() {
...
 
-        int dbSize = locationService.findAll().size();
-        Location updated = importService.createOrUpdateLocation(newLocation, EXISTING_EXTERNAL_ID, PROVIDER);
+        location.setIsConfidential(true);
+        location.setName(NEW_STRING);
+        var updated = importManager.createOrUpdateLocation(location, EXISTING_EXTERNAL_ID, PROVIDER);
 
-        assertEquals(dbSize, locationService.findAll().size());
+        assertEquals(1, locationService.findAll().size());
+        assertEquals(EXISTING_STRING, locationService.findAll().get(0).getName());
```
wojtekbu: So when a location is updated with confidential the old one is still available for other providers and the new one for the owner? In my opinion, this is misleading because the location is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -109,142 +172,113 @@ public void shouldNotCreateLocationIfConfidential() {
...
-        PhysicalAddress updated = importService.createOrUpdatePhysicalAddress(newAddress, location);
+        PhysicalAddress newAddress = new PhysicalAddress().address1(NEW_STRING).city(NEW_STRING).stateProvince(NEW_STRING);
+        newAddress.setIsConfidential(true);
+        location.setPhysicalAddress(newAddress);
+        locationImportService.createOrUpdateLocation(location, EXISTING_EXTERNAL_ID, PROVIDER);
 
-        assertEquals(dbSize, locationService.findAll().size());
-        assertNull(updated);
+        assertEquals(1, physicalAddressService.findAll().size());
+        assertEquals(EXISTING_STRING, physicalAddressService.findAll().get(0).getAddress1());
```
wojtekbu: In my opinion, this is misleading because the physical address is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -109,142 +172,113 @@ public void shouldNotCreateLocationIfConfidential() {
...
-        PostalAddress updated = importService.createOrUpdatePostalAddress(newAddress, location);
+        PostalAddress newAddress = new PostalAddress().address1(NEW_STRING).city(NEW_STRING).stateProvince(NEW_STRING);
+        newAddress.setIsConfidential(true);
+        location.setPostalAddress(newAddress);
+        locationImportService.createOrUpdateLocation(location, EXISTING_EXTERNAL_ID, PROVIDER);
 
-        assertEquals(dbSize, locationService.findAll().size());
-        assertNull(updated);
+        assertEquals(1, postalAddressService.findAll().size());
+        assertEquals(EXISTING_STRING, postalAddressService.findAll().get(0).getAddress1());
```
wojtekbu: In my opinion, this is misleading because the postal address is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -109,142 +172,113 @@ public void shouldNotCreateLocationIfConfidential() {
...
-        AccessibilityForDisabilities updated = importService.createOrUpdateAccessibility(newAccessibility, location);
+        AccessibilityForDisabilities newAccessibility = new AccessibilityForDisabilities().accessibility(NEW_STRING).details(NEW_STRING);
+        newAccessibility.setIsConfidential(true);
+        location.setAccessibilities(Set.of(newAccessibility));
+        locationImportService.createOrUpdateLocation(location, EXISTING_EXTERNAL_ID, PROVIDER);
 
-        assertEquals(dbSize, locationService.findAll().size());
-        assertNull(updated);
+        assertEquals(1, accessibilityService.findAll().size());
+        assertEquals(OTHER_STRING, accessibilityService.findAll().get(0).getAccessibility());
```
wojtekbu: In my opinion, this is misleading because the accessibility is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -253,25 +287,29 @@ public void shouldNotCreateOrganizationIfConfidential() {
...
-        int dbSize = organizationService.findAllDTOs().size();
-        Organization updated = importService.createOrUpdateOrganization(newOrganization, EXISTING_EXTERNAL_ID, PROVIDER,
+        newOrganization.setIsConfidential(true);
+        newOrganization.setName(NEW_STRING);
+        var updated = importManager.createOrUpdateOrganization(newOrganization, EXISTING_EXTERNAL_ID, PROVIDER,
             new DataImportReport());
 
-        assertEquals(dbSize, organizationService.findAllDTOs().size());
+        assertEquals(1, organizationService.findAllDTOs().size());
+        assertEquals(EXISTING_STRING, organizationService.findAllDTOs().get(0).getName());
```
wojtekbu: In my opinion, this is misleading because the organization is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -295,213 +333,159 @@ public void shouldNotUpdateServiceIfConfidential() {
...
 
-        Phone newPhone = new Phone().number(NEW_STRING).srvc(service);
-        newPhone.setIsConfidential(true);
-
-        importService.createOrUpdatePhonesForService(Set.of(newPhone), service, null);
+        importManager.createOrUpdateService(service, EXISTING_EXTERNAL_ID, PROVIDER, new DataImportReport());
 
-        assertEquals(0, phoneService.findAll().size());
+        assertEquals(1, phoneService.findAll().size());
+        assertEquals(EXISTING_STRING, phoneService.findAll().get(0).getNumber());
```
wojtekbu: In my opinion, this is misleading because the phone is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -295,213 +333,159 @@ public void shouldNotUpdateServiceIfConfidential() {
...
 
-        int dbSize = eligibilityService.findAll().size();
-        Service serviceFromDb = serviceService.findWithEagerAssociations(EXISTING_EXTERNAL_ID, PROVIDER).get();
-        Eligibility updated = importService.createOrUpdateEligibility(newEligibility, serviceFromDb);
+        importManager.createOrUpdateService(service, EXISTING_EXTERNAL_ID, PROVIDER, new DataImportReport());
 
-        assertEquals(dbSize, eligibilityService.findAll().size());
-        assertNull(updated);
+        assertEquals(1, eligibilityService.findAll().size());
+        assertEquals(EXISTING_STRING, eligibilityService.findAll().get(0).getEligibility());
```
wojtekbu: In my opinion, this is misleading because the eligibility is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
```diff
@@ -295,213 +333,159 @@ public void shouldNotUpdateServiceIfConfidential() {
...
-        languageService.findAll().size();
-        importService.createOrUpdateLangsForService(Set.of(newLanguage), service, null);
+        Language otherLanguage = new Language().language(OTHER_STRING);
+        otherLanguage.setIsConfidential(true);
+        service.setLangs(Set.of(otherLanguage));
+        importManager.createOrUpdateService(service, EXISTING_EXTERNAL_ID, PROVIDER, new DataImportReport());
 
-        assertEquals(0, languageService.findAll().size());
+        assertEquals(1, languageService.findAll().size());
+        assertEquals(EXISTING_STRING, languageService.findAll().get(0).getLanguage());
```
wojtekbu: In my opinion, this is misleading because the lang is outdated.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: The same physical location for two different providers will be available as two separate entities in the DB. Currently, we're only removing some detached (historical) entities, but we should consider the behavior more accurately for all of them in the future. For now, only entities with OneToMany relations to them are removed when are no longer connected with no other. Those are Languages, Programs, OpeningHours, Contacts, and Phones.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: same here. I agree, but we probably have to create a separate a task for it, as it's working the same way as before changing the ImportService.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: same here. I agree, but we probably have to create a separate a task for it, as it's working the same way as before changing the ImportService.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: same here. I agree, but we probably have to create a separate a task for it, as it's working the same way as before changing the ImportService.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: same here. I agree, but we probably have to create a separate a task for it, as it's working the same way as before changing the ImportService.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: same here. I agree, but we probably have to create a separate a task for it, as it's working the same way as before changing the ImportService.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: same here. I agree, but we probably have to create a separate a task for it, as it's working the same way as before changing the ImportService.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: You're right in this case, it shouldn't happened.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
oskarhinc: You're right in this case, it shouldn't happened.

src/test/java/org/benetech/servicenet/service/impl/ImportServiceConfidentialityImplIntTest.java
wojtekbu: @oskarhinc can you create that task?

oskarhinc: Sure. There is something similar in the backlog so I've added a new comment there: https://github.com/benetech/ServiceNet/issues/220

Issue #261, opened 2019-02-22
Optimze location matching algorithm


oskarhinc:
Request to the Google API are done too often, so the location matching is taking too long. Additionally,  the location matching might be the cause of problems with the heap size. The algorithm should be optimized.

oskarhinc:
@wojtekbu after the latest improvements, the algorithm should work much faster. Please review: https://github.com/benetech/ServiceNet/pull/278

wojtekbu:
@oskarhinc works fine

PR #268, opened 2019-02-23
 #261: Saving geocoding result while importing location

oskarhinc:


src/main/java/org/benetech/servicenet/service/impl/GeocodingResultServiceImpl.java
```diff
@@ -58,4 +72,30 @@ public void delete(UUID id) {
...
     }
+
+    @Override
+    public List<GeocodingResult> createOrUpdateGeocodingResult(Location location) {
+        if (location.getPhysicalAddress() == null) {
+            return new ArrayList<>();
+        }
+
+        String addressString = geoApi.extractAddressString(location.getPhysicalAddress());
+        List<GeocodingResult> currentResults = geocodingResultRepository.findAllByAddress(addressString);
```
wojtekbu: I recommend adding an index on that. Since it is a clob you should narrow an indexed part of an address to for example 50 chars. I think that coverage of address.getAddress1(), address.getCity() would be enough to narrow search results. loner string would be just too much computation.

src/main/java/org/benetech/servicenet/domain/GeocodingResult.java
```diff
@@ -55,7 +58,23 @@ public GeocodingResult longitude(Double longitude) {
         return this;
     }
 
-    // jhipster-needle-entity-add-getters-setters - JHipster will add getters and setters here, do not remove
```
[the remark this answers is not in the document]
oskarhinc: I left both

src/main/java/org/benetech/servicenet/service/impl/GeocodingResultServiceImpl.java
oskarhinc: Sure, nice catch

oskarhinc: Unfortunatelly, as the address is of type CLOB, neither h2 or psql does not support it fully. We can create a walkaround later if it will be too slow. @wojtekbu what do you think?

wojtekbu: Works for me

PR #278, opened 2019-02-25
#261: Improved location matching algorithm

oskarhinc:


src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
```diff
@@ -34,19 +38,55 @@
...
+        }
+    }
+
+    private float countRatioIfCoordinatesAreMissing(Location location1, Location location2, MatchingContext context) {
         float max = NO_MATCH_RATIO;
-        for (GeocodingResult result1 : geocodingResultService.findAllForLocationOrFetchIfEmpty(location1, context)) {
-            for (GeocodingResult result2 : geocodingResultService.findAllForLocationOrFetchIfEmpty(location2, context)) {
+
+        for (GeocodingResult result1 : getGeoCodingResult(location1, context)) {
+            for (GeocodingResult result2 : getGeoCodingResult(location2, context)) {
```
wojtekbu: How many API calls can be executed in the worst scenario?

src/main/java/org/benetech/servicenet/service/impl/GeocodingResultServiceImpl.java
```diff
@@ -71,12 +71,31 @@ public void delete(UUID id) {
     }
 
     @Override
-    public List<GeocodingResult> createOrUpdateGeocodingResult(Location location, MatchingContext context) {
+    public Optional<GeocodingResult> getGeocodeForLocationIfUnique(Location location, MatchingContext context) {
```
wojtekbu: Why is this optional when we can still have multiple geolocations for one location?

src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
oskarhinc: two, if somehow the result was not saved in the db during the import. However it should be saved, so We're searching the db here. In fact, I should not search it in the nested loop, so I'll extract it.

src/main/java/org/benetech/servicenet/service/impl/GeocodingResultServiceImpl.java
oskarhinc: Because we can't store the coordinates in location if there were multiple responses from the Google API. In fact, Maybe we should store them at all, as it might confuse the partners that there is a data they didn't provide.

src/main/java/org/benetech/servicenet/domain/GeocodingResult.java
```diff
@@ -27,10 +25,8 @@
 
     private static final long serialVersionUID = 1L;
 
-    @Lob
     @NotNull
-    @Type(type = "org.hibernate.type.TextType")
-    @Column(name = "address", columnDefinition = "clob", nullable = false)
```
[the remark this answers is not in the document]
oskarhinc: it's the default value, and I think it's not necessary then

Issue #75, opened 2018-12-28
Add custom bean validator before saving mapped data


oskarhinc:
As the amount of data from partners is going to be really huge, it is required to validate mapped entities before saving them. Otherwise, none of the data will be persisted, because there was a DB exception with a single entity. Doing it manually with each entity is not the best solution, so we should use something like Spring Bean Validation to handle this. It would be nice to have details about beans that were validated with errors stored in the database, so it could be verified and fixed then, but fixing those errors should be a separate feature. 

This should be done after #176

wojtekbu:
@oskarhinc  works fine except that error messages are unformatted. For example 'nullCannot connect with Google Maps API Cannot connect with Google Maps API'

oskarhinc:
Nice natch @wojtekbu . I've improved the reports handling for errors as well now: https://github.com/benetech/ServiceNet/pull/279

PR #279, opened 2019-02-25
#75: Provide better error handlig for reports

oskarhinc:


src/main/java/org/benetech/servicenet/builder/ReportErrorMessageBuilder.java
```diff
@@ -29,6 +30,17 @@
         return builder.toString();
     }
 
+    public static String buildForError(String message, DataImportReport report) {
+        String prevPart;
+        if (StringUtils.isBlank(report.getErrorMessage())) {
+            prevPart = FIRST_LINE;
```
wojtekbu: I think blank errors shouldn't be omitted but marked as undefined

oskarhinc: It's not handling blank errors but rather current errorMessage for the report, which will be null for the first error

Issue #45, opened 2018-12-10
Add searching activities to the main page


wojtekbu:
Add a list of resources with related conflicts to the main page
The parent ticket can be found here: https://github.com/benetech/ServiceNet/issues/11
According to the mock https://user-images.githubusercontent.com/12927553/48443971-55f41d00-e758-11e8-96ad-6596c7a6092c.png add searching activities to the main page.

johnhbenetech:
As a result of the additional stabilization work we are performing, I have reprioritized this above filtering and location conflict detection. 

Once the stabilization of production/master branch is finished, then exploration of rollbacks, geocoding, and sorting are completed in the dev branch - we should focus on this next.

I think at first I would love a quick implementation that does Organization Name only (case insensitive, partial matches included).

After this is on dev, we can revisit later to include other fields and similar strings with low edit distances.

oskarhinc:
@wojtekbu could you test it after the review of https://github.com/benetech/ServiceNet/pull/284, please?

johnhbenetech:
Can we add a clear search element to the front end? something like an 'x' in a circle in the end of the search bar?

wojtekbu:
Works fine:

- searching using organizations' name,
- partial search,
- case insensitive search,
- total number of activities,
- clear searching,
- sorting with search,
- sorting without search,

PR #284, opened 2019-02-28
#45: Added searching on backend with full connection to the front end

oskarhinc:


johnhbenetech:
can we add a clear search element to the front end, something like an 'x' in a circel in the end of the search bar?

oskarhinc:
@johnhbenetech sure

Issue #200, opened 2019-02-06
Add all available information to side-by-side comparison view


johnhbenetech:
We need to allow the side by side view to display all available information for each partner's record (even if that entity isnt available for both partners). Right now it seems to be only a select few fields based on our ability to detect conflicts.

johnhbenetech:
I'm moving this priority up based on some early user testing. I like how the single record view looks (how it includes all the entities) we should have this same type of view for all side by side records.

oskarhinc:
@wojtekbu Could you review https://github.com/benetech/ServiceNet/pull/285 and then test this task, please?

johnhbenetech:
@oskarhinc @wojtekbu Please review #286

wojtekbu:
Works fine

PR #285, opened 2019-02-28
#200: Added all details to the multiple-record-view

oskarhinc:


src/main/webapp/app/modules/conflicts/multiple/multiple-record-view.tsx
```diff
@@ -69,7 +69,7 @@ export class MultipleRecordView extends React.Component<IMultipleRecordViewProp,
               <h4 className="from">
                 <Translate contentKey="multiRecordView.yourData" />
               </h4>
-              <Details organization={baseRecord.record.organization} {...this.props} exclusions={baseRecord.record.exclusions} />
+              <Details activity={baseRecord} {...this.props} exclusions={baseRecord.record.exclusions} isBaseRecord />
```
wojtekbu: Why 'isBaseRecord' default value is set to true? There will be always more children.

src/main/webapp/app/modules/conflicts/single/details.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const sideSection = firstMatch ? (
+      <Col sm="3">
+        <Jumbotron className="jumbotron">
+          <div className="jumbotron-header">
+            <h4 className="jumbotron-header-title">
+              <Translate contentKey="singleRecordView.details.compare.title" />
+            </h4>
+            <p className="jumbotron-header-text">
+              {activity.organizationMatches.length}
+              {activity.organizationMatches.length === 1 ? (
```
wojtekbu: Why do you need this upper {activity.organizationMatches.length} ?

src/main/webapp/app/modules/conflicts/single/details.tsx
```diff
@@ -0,0 +1,99 @@
...
+                  ? `${firstMatch.partnerVersionName}`
+                  : `${firstMatch.partnerVersionName} +${activity.organizationMatches.length - 1}`}
+              </Link>
+            </div>
+          </div>
+        </Jumbotron>
+      </Col>
+    ) : null;
+
+    const columnSize = 6;
```
wojtekbu: please make it UPPERCASE

src/main/webapp/app/modules/conflicts/multiple/multiple-record-view.tsx
oskarhinc: It's for the multi-record-view to indicate how some elements should be visible

src/main/webapp/app/modules/conflicts/single/details.tsx
oskarhinc: It's not global const value but rather a local one, but it won't be changed so tsLisnt suggests making it 'const'

src/main/webapp/app/modules/conflicts/single/details.tsx
oskarhinc: To indicate the number of partners that have similar records

Issue #291, opened 2019-03-08
Updating data causes several issues that cause fatal errors


johnhbenetech:
Steps to reproduce:
* Start from an empty database
* Take a recent copy of SMC data and import, wait for import process to finish
* Take a recent copy of LAAC data and import, wait for import process to finish, wait for conflict detection to finish

* Make smc user, Log into smc user, note a single matching organization with conflicts, note the name of the matching record for LAAC. Note a specific conflict (such as URL) and note the last updated timestamp

* make LAAC user and log in. See that records display and platform roughly works

* Go back to the LAAC spreadsheet. Find that record that matched above. Update the field involved in the conflict that you noted above (such as URL) to something else.

* Reupload the LAAC data to the platform, wait for processing to complete, wait for conflicts to finish

ISSUE 1: log in to smcuser. find the record from above. Note: date update field has not changed to reflect last update time (from the second upload).

ISSUE 2: From issue 1, conflicts have not updated to reflect new data. Existing conflicts still represent old data value (old URL value, old date updated value). Even looking at direct query of conflict db reveals no new conflicts being created for this record. Looking at side-by-side view, however, the data has been updated for the matching record to reflect the new value (even though a conflict hasnt)

ISSUE 3: logout, and log back in as laacuser. A fatal ISE occurs rendering the platform useless for this user. The error seems to reference being unable to find a reference (even though everything worked after the first upload). I have to assume then that the platform is expecting something from the first version of the record that no longer exists. I haven't gone as deep as to ensure each entity's ID remains consistent across updates (and maybe it shouldn't if the approach is just to make a new entity). But the log below seems to indicate that maybe a Contact entity couldnt be found. There was no more log activity following that so I think it killed any subsequent attempt to display records.

javax.persistence.EntityNotFoundException: Unable to find org.benetech.servicenet.domain.Contact with id ce59cfb3-eed7-4610-afc2-dafdd874cc8e

wojtekbu:
Hi, @johnhbenetech I figured it out how to handle ISSUE 3. I'm working on two others, but I want to make sure that we both have the same organization-match-threshold value, which is 1.0 in my env.

johnhbenetech:
@wojtekbu yes I'm using 1.0. I did set the always location option to false, because I was having performance issues when it was on

johnhbenetech:
@wojtekbu @oskarhinc This still isn't working as intended.

[...]

PR #338, opened 2019-03-19
#291: Added updating records only when record has changed

wojtekbu:


src/main/java/org/benetech/servicenet/conflict/detector/AbstractDetector.java
```diff
@@ -17,20 +17,28 @@ protected boolean areEquals(String current, String offered) {
...
     protected List<Conflict> detectConflicts(T current, String name, String name2, String fieldName) {
         List<Conflict> conflicts = new LinkedList<>();
-        if (!this.areEquals(name, name2)) {
+        if (detect(name, name2)) {
             conflicts.add(createConflict(current, name, name2, fieldName));
         }
         return conflicts;
     }
 
+    protected boolean detect(String name, String name2) {
```
oskarhinc: this method name does not indicate what it does, and it might lead to confusions in the future

oskarhinc:
Despite this one comment, I see no issues here so I if you're not able to fix it today, let me know as it can be adjusted later

src/main/java/org/benetech/servicenet/conflict/detector/AbstractDetector.java
wojtekbu: this is logic is responsible for detecting conflicts in a method called detectConflicts, but this one does not return conflicts as a result. That's why I called it detect

oskarhinc: it's a boolean method and it's not obvious what the result will be

wojtekbu:
> Despite this one comment, I see no issues here so I if you're not able to fix it today, let me know as it can be adjusted later

Ok, I'll fix that later

Issue #428, opened 2019-05-23
Create Shelter Entity


johnhbenetech:
Create a new entity called Shelter containing the following fields:

Any fields designated as text should not have length limits. Ignore the "D"/"M" for now, I'm using that as a notation that will become relevant in a later task.

General Info
* Agency Name (text) M
* Program Name (text) M
* Alternate Name (text) D

More Info
* Website (text) D

Eligibility
* Eligibility Details (text) M
* Documents Required (text) M
* Language (multi select) D
- English, Spanish, Russian, Hmong, Vietnamese, Tagalog, Chinese

Process
* Application Process (text) M
* Fees (text) D
* Defined Coverage Area (multi select - counties) D
-  Sacramento, Yolo, Solano, Contra Costa, Sutter, Placer, El Dorado, San Joaquin, Amador

Hours
* Program Hours (text) D
* Holiday Schedule (text) D

Contact
* Email (text)(many to one) M

* Phone Entity (many to one) M
** Phone Number (text) (for each phone) M
** Phone Type (text)(for each phone) M

Location (single) 
* Address 1 (text) M
* Address 2 (text) M
* City (text) M
* Zipcode (text) M
* Location Description (text) D
* Bus service (text) D
* Transportation (text) D
* Disability Access (text) D

Beds
* Number of available beds (number) M
* Waitlist (number) M
* Last updated date (automatic)(for both beds and waitlist) M

Tags (multi select) D, Options as followed:
* Audience Men
* Audience Women
* Audience Families
* Accessible
* Accepts Pets
* Accept personal belongings
* Provide medical beds

johnhbenetech:
@jkondrat Here's a good first task for beginning work on the new project. Please note there are a few intricacies that I'll need some suggestions from you on how to implement.

1. Multi-Select fields. I can imagine doing this by creating a separate entity for each of these fields (with a key back to the shelter entity), but that approach seems a little clunky. How do you think we should handle this? Maybe storing as a pipe-separated value and doing the rendering/validation in the frontend. That seems ugly too though, let me know what you think with the understanding that at some point we will want to display all of the above fields on a single page.

2. Phones. Many Phones can be associated with a Shelter. For each Phone, there are two related text fields - Phone Number, and Phone Type. Once again not sure how to handle this - we will need to display all of these fields on the same page.

[...]

PR #470, opened 2019-06-05
#428: Create Shelter Entity

jkondrat:


src/main/java/org/benetech/servicenet/service/impl/ShelterServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+        if (beds != null) {
+            beds.setShelter(shelter);
+            shelter.setBeds(bedsRepository.save(beds));
+        }
+
+        Set<Phone> phones = shelter.getPhones().stream()
+            .filter(phone -> phone.getNumber() != null)
+            .collect(Collectors.toSet());
+        for (Phone phone:
+        phones) {
```
pmuchowski: this should be in one line

src/main/java/org/benetech/servicenet/service/impl/ShelterServiceImpl.java
```diff
@@ -0,0 +1,123 @@
...
+            phone.setShelter(shelter);
+        }
+
+        shelter.setPhones(phones);
+        List<String> emails = shelter.getEmails().stream()
+            .filter(email -> email != null && !email.isEmpty())
```
pmuchowski: you can use StringUtils.isNotBlank

src/main/webapp/i18n/en/language.json
```diff
@@ -1,23 +1,14 @@
 {
   "serviceNetApp": {
-    "language": {
```
pmuchowski: why did you removed those messages? are you sure it's not used anywhere?

tslint.json
```diff
@@ -97,6 +97,7 @@
     "space-in-parens": [true, "never"]
   },
   "rules": {
+    "jsx-no-lambda": false,
```
pmuchowski: why did you disabled this rule?

Issue #339, opened 2019-03-19
ability to dismiss record when not good match


CloVasco:
Once red button removed - Include a way for record to disappear after users take action on it if not good match
- what happens when click on the red button? disappear automatically - yes. new window to confirm - yes.
- for that user only? or for the team?  - Clotilde to ask partners
-at what level - the record or service/address? after discussion on 3/27 - decision to do by whole record. 
- what do we do when dismissed?  can the user see it again or never ever - no. John will review after

CloVasco:
@laurabenetech and @johnhbenetech 
Our partners voted on this item ;)! See detailed notes in the attached doc. In a nutshell:
-5/6 said it should be for the whole organization
-2 raised the idea that they want to still have access to it if needed - a queue where can go back, look at it and reintegrate it if we want (UWBA) // a note on top of the side by side page where the bad matches are mentioned (Eden)
-2 raised the idea that it's not a black and white decision - could give more details on why it's not a match (LAAC)/Benetech should clarify and define what is a match or not - otherwise won't be consistent between users and orgs (HL)

[dismissal when not a good match - for whom.docx](https://github.com/benetech/ServiceNet/files/3036390/dismissal.when.not.a.good.match.-.for.whom.docx)

CloVasco:
@laurabenetech and @johnhbenetech 
So my proposition:
-	The decision is for everybody in the organization
-	We still give them the ability to go back to the wrong match and reintegrate it 
o	I like Rick’s idea of adding a note somewhere in the side by side view to show that some other records have been dismissed and when you click on it, it shows dismissed records (+who and when dismissed them) and you can reinstate them
-	Open questions
o	Do we want to give them ability to give explanation? I could argue both ways
o	Should we clarify what a good/bad match is? Yes probably but it’s a complex discussion

CloVasco:
Do you want to do a back and forth through github or have a quick phone call to decide before Laura works on visuals? If you think it will be faster to reach a decision, I can talk on:
-	Wednesday – at 10am or after 3.320
-	Thursday – between 11-1pm and 3pm

johnhbenetech:
Proposition seems fine. I'll just need a visual for how to display the "dismissed records" and bringing them back.

On your open questions, technically there isn't a big difference. I think the question is would anyone actually fill it out and how would we use that info. We can always make it optional.

[...]

PR #477, opened 2019-06-06
#339: Ability to dismiss record when not good match

pmuchowski:


src/main/webapp/app/modules/conflicts/dismissed/dismissed-match-view.tsx
```diff
@@ -0,0 +1,90 @@
...
+  componentDidMount() {
+    this.props.getBaseRecord(this.props.orgId);
+    this.props.getPartnerRecord(this.props.partnerId);
+  }
+
+  render() {
+    const { baseRecord, partnerRecord, orgId } = this.props;
+    const loading = (
+      <Col>
+        <h2>Loading...</h2>
```
awalkowiak: I guess we could add "Loading..." in translations

src/main/webapp/app/modules/conflicts/dismissed/dismissed-matches.tsx
```diff
@@ -0,0 +1,120 @@
...
+        toast.error(translate('dismissedMatches.reinstateError'));
+      });
+  };
+
+  render() {
+    const { dismissedMatches, orgId } = this.props;
+    return (
+      <div>
+        <Link to={`/${this.props.matches.length > 0 ? 'multi' : 'single'}-record-view/${orgId}`}>
+          <FontAwesomeIcon icon="arrow-left" />{' '}
```
awalkowiak: \&nbsp; looks prettier

Issue #475, opened 2019-06-05
Improve URL conflict detection


johnhbenetech:
Currently conflicts are logged for URL even when there are no meaningful differences between values - such as:

* http vs no http
* www vs no www
* http vs https
* trailing / vs no trailing /

Make URL conflict detection smarter so it only detects meaningful difference.

For right now we should still log a conflict for different directory/pages on the same domain, e.g.:
* google.com vs google.com/page.html
* google.com/dir vs google.com/example

awalkowiak:
Dev note: Wwe could use regex groups to cleanup site and compare only that part

bognasoldev89:
Unfortunately, this issue needs more work, I found a few examples like this one:
![Uploading Screenshot from 2019-06-14 14-46-00.png…]()

pmuchowski:
@bognasoldev89 did you uploaded this organization or just looked through the existing records?
I'm not sure if you aware of this, but you need to update an organization for it's conflicts to be updated. There might still exist some old incorrect URL conflicts, because those was detected before this code was merged.

bognasoldev89:
@pmuchowski I was not aware actually, I am gonna check it further then. Thanks for the advice.

bognasoldev89:
After uploading data I've found that after uploading data as Healthleads (differences mentioned in the first comment) the app still sees conflicts. I've also tried ShelterTech data and I've found that some of these works (Trauma Recovery Center).

bognasoldev89:
Confirmed that URL conflict detection is improved  and ignores not meaningful differences between values

kkaczmarczyk:
The solution is not null-safe. Moving ticket back.

![image](https://user-images.githubusercontent.com/3207119/60607393-77744880-9dbd-11e9-944c-84a69d7a3d40.png)

kkaczmarczyk:
@mariaogryczak make sure to include a test for null case please.

bognasoldev89:
Confirmed that URL conflict detection is improved, ignores not meaningful differences between values and is null-safe.

PR #514, opened 2019-06-25
#475: Improve URL conflict detection

mariaogryczak:


src/main/java/org/benetech/servicenet/matching/counter/UrlSimilarityCounter.java
```diff
@@ -14,48 +13,38 @@
...
-    
-    @Value("${similarity-ratio.weight.url.equal-upper-cased}")
-    private float uppercasedWeight;
 
     @Override
     public float countSimilarityRatio(String url1, String url2, MatchingContext context) {
         if (StringUtils.isBlank(url1) || StringUtils.isBlank(url2)) {
             return NO_MATCH_RATIO;
         }
-        if (areNormalizedAndUpperCasedDifferent(url1, url2)) {
```
pmuchowski: this is used to calculate the match, you can't make this changes, because it will affect the matching algorithm.
revert all of the changes in this file

src/main/java/org/benetech/servicenet/conflict/detector/AbstractDetector.java
```diff
@@ -4,17 +4,27 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
 
 import java.time.ZonedDateTime;
 import java.util.LinkedList;
 import java.util.List;
 
 public abstract class AbstractDetector<T extends AbstractEntity> {
 
+    @Autowired
+    private UrlSimilarityCounter urlSimilarityCounter;
```
pmuchowski: you should not change the AbstractDetector, make the changes in OrganizationConflictDetector instead

src/main/java/org/benetech/servicenet/matching/counter/UrlSimilarityCounter.java
```diff
@@ -14,48 +13,38 @@
...
-            return uppercasedWeight;
+        if (areNormalizedAndUpperCasedTheSame(url1, url2)) {
+            return COMPLETE_MATCH_RATIO;
         }
 
-        return COMPLETE_MATCH_RATIO;
+        return NO_MATCH_RATIO;
     }
 
     private String normalize(String url) {
```
pmuchowski: I think you can move the normalization code to some separate utils class and use it in OrganizationConflictDetector too, you can name it NormalizationUtils or something like that

src/main/java/org/benetech/servicenet/conflict/detector/OrganizationConflictDetector.java
```diff
@@ -32,3 +34,5 @@
         return conflicts;
     }
 }
+
+
```
pmuchowski: remove the "extra" empty lines

src/main/java/org/benetech/servicenet/util/UrlNormalizationUtils.java
```diff
@@ -0,0 +1,32 @@
+package org.benetech.servicenet.util;
+
+import java.util.Locale;
+
+public final class UrlNormalizationUtils {
+
+    public static final String WWW = "WWW.";
```
pmuchowski: those can be private

src/main/java/org/benetech/servicenet/matching/counter/UrlSimilarityCounter.java
```diff
@@ -26,36 +22,19 @@ public float countSimilarityRatio(String url1, String url2, MatchingContext cont
...
-        }
-        if (result.toUpperCase(Locale.ROOT).endsWith(TRAILING)) {
-            result = result.substring(0, result.length() - 1);
-        }
-        return result;
-    }
-
     private boolean areNormalizedAndUpperCasedDifferent(String url1, String url2) {
-        return !normalize(url1).toUpperCase(Locale.ROOT).equals(normalize(url2).toUpperCase(Locale.ROOT));
+        return !UrlNormalizationUtils.normalize(url1).toUpperCase(Locale.ROOT).
```
pmuchowski: The dot should be in the new line,
run the tests "./mvnw test", the checkstyle will fail

Issue #367, opened 2019-04-05
Add "go back to top" button


CloVasco:
Add a button to go back to top of the page on pages:
- single record view
-side by side view

CloVasco:
@laurabenetech could you please provide the visual? thanks in advance

laurabenetech:
@CloVasco 
The Back to Top button appears in the lower right-hand corner of the screen after the user has scrolled downwards so that they no longer see the top of the screen (Org Details section).

<img width="94" alt="Back to Top button" src="https://user-images.githubusercontent.com/48697326/57555341-e044c480-7328-11e9-8352-f348a8241363.png">

Once the user has scrolled down or is scrolling, the button persists in the bottom righthand corner of the screen so that the user can easily click the button anytime to return to the top. Once user clicks the button and returns to the top, the button disappears.

@johnhbenetech  I’m not sure how to technically describe when this should appear. If it makes sense to do this in terms of number of seconds of downward scrolling or after the user has reached the point where the Locations section is at the top of the screen, we can do that. I just want to ensure that the button doesn’t appear immediately if users do a small amount of scrolling (but are still viewing that first screen of info—the Org Details).

CloVasco:
@laurabenetech 
thanks. Over to you, @johnhbenetech

mariaogryczak:
@johnhbenetech 
Do you have any specific requirements connected with this button? I mean:
after how much scrolling should it appear?
how fast should it scroll up to the top?

Moreover, is the "side by side view" the "multi record view" or is it something else?

johnhbenetech:
@mariaogryczak 

Yes side-by-side = multi record view

With regards to behavior. This library seems to perform how I would want it:
https://www.npmjs.com/package/react-scroll-up-button

Button appears after a small amount of scrolling. Clicking button smooth scrolls back to top. We would just need to replace the button with the design from above.

bognasoldev89:
Confirmed that button exists and works perfectly fine

PR #515, opened 2019-06-26
#367: Add 'go back to top' button

mariaogryczak:


src/main/webapp/app/modules/conflicts/shared/components/additional-details.tsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import '../shared-record-view.scss';
 import { connect } from 'react-redux';
+import ScrollToTop from 'react-scroll-up';
```
pmuchowski: did you added as a new dependency in package.json?

pmuchowski:
@mariaogryczak remove the url conflict detection commit from this PR
When you creating a new branch you need to checkout to the develop first, otherwise all of the changes from the current branch will be moved to the new branch

src/main/webapp/app/modules/conflicts/shared/components/additional-details.tsx
mariaogryczak: Before you wrote about it - no. Now I fixed it. 
I have one more question:
I also added an image of a button "content/images/backToTop_transparent_resized.png". Does it update automatically during commit or I have to add it in a different way?

src/main/webapp/app/modules/conflicts/shared/components/additional-details.tsx
```diff
@@ -59,6 +60,11 @@ export class AdditionalDetails extends React.Component<IAdditionalDetailsProp, I
           ))}
           {additionalFields}
         </Collapse>
+        <div>
+          <ScrollToTop showUnder={150} duration={500} style={{ display: 'inline-block', marginRight: '10px' }}>
+            <FontAwesomeIcon size="3x" icon={'angle-up'} color="lightblue" />
+          </ScrollToTop>
```
pmuchowski: I think you can add the "Back to Top" text under the icon too

pmuchowski:
@mariaogryczak there is still the "Improve URL conflict detection" commit in this PR, remove it

Issue #456, opened 2019-05-29
Fix errors after user logs out


johnhbenetech:
After some recent change, after a user logs out they see several error pop ups relating to full auth requires to access this resource.

We need to either fix the underlying issue or suppress these errors. I hope this is quick, but if not let me know.

johnhbenetech:
errors from a single log out

2019-05-29 23:46:55.362  WARN 8 --- [  XNIO-2 task-2] o.z.problem.spring.common.AdviceTrait    : Unauthorized: Full authentication is required to access this resource
2019-05-29 23:46:55.363  WARN 8 --- [  XNIO-2 task-5] o.z.problem.spring.common.AdviceTrait    : Unauthorized: Full authentication is required to access this resource
2019-05-29 23:46:55.364  WARN 8 --- [  XNIO-2 task-8] o.z.problem.spring.common.AdviceTrait    : Unauthorized: Full authentication is required to access this resource
2019-05-29 23:46:55.369  WARN 8 --- [ XNIO-2 task-11] o.z.problem.spring.common.AdviceTrait    : Unauthorized: Full authentication is required to access this resource
2019-05-29 23:46:55.374  WARN 8 --- [ XNIO-2 task-11] .m.m.a.ExceptionHandlerExceptionResolver : Resolved exception caused by handler execution: org.springframework.security.authentication.InsufficientAuthenticationException: Full authentication is required to access this resource
2019-05-29 23:46:55.374  WARN 8 --- [  XNIO-2 task-5] .m.m.a.ExceptionHandlerExceptionResolver : Resolved exception caused by handler execution: org.springframework.security.authentication.InsufficientAuthenticationException: Full authentication is required to access this resource
2019-05-29 23:46:55.374  WARN 8 --- [  XNIO-2 task-2] .m.m.a.ExceptionHandlerExceptionResolver : Resolved exception caused by handler execution: org.springframework.security.authentication.InsufficientAuthenticationException: Full authentication is required to access this resource
2019-05-29 23:46:55.374  WARN 8 --- [  XNIO-2 task-8] .m.m.a.ExceptionHandlerExceptionResolver : Resolved exception caused by handler execution: org.springframework.security.authentication.InsufficientAuthenticationException: Full authentication is required to access this resource
2019-05-29 23:46:55.476  WARN 8 --- [ XNIO-2 task-13] o.z.problem.spring.common.AdviceTrait    : Unauthorized: Full authentication is required to access this resource
[...]

PR #516, opened 2019-06-26
#456: Fix errors after user logs out

kkaczmarczyk:


src/main/webapp/app/shared/reducers/authentication.ts
```diff
@@ -64,7 +70,8 @@ export default (state: AuthenticationState = initialState, action): Authenticati
     case ACTION_TYPES.LOGOUT:
       return {
         ...initialState,
-        showModalLogin: true
+        showModalLogin: true,
+        loggingOut: false
```
pmuchowski: I think you can also set it to false in the action above (login success) just to be safe

src/main/webapp/app/shared/reducers/authentication.ts
```diff
@@ -32,6 +33,11 @@ export type AuthenticationState = Readonly<typeof initialState>;
 export default (state: AuthenticationState = initialState, action): AuthenticationState => {
   switch (action.type) {
     case REQUEST(ACTION_TYPES.LOGIN):
+    case REQUEST(ACTION_TYPES.LOGOUT):
+      return {
+        ...state,
+        loggingOut: true
```
pmuchowski: why it's set to true on login request?

kkaczmarczyk: fixeed

src/main/webapp/app/shared/reducers/authentication.ts
kkaczmarczyk: ok

kkaczmarczyk:
additionally I fixed case for LOGIN action success that seemed to be broken

Issue #147, opened 2019-01-22
Extend DataImportReport in order to log failed parsing attempts


wojtekbu:
As soon as https://github.com/benetech/ServiceNet/issues/75 will be done we can log all failed attempts to the upload reports. Those logs should point out which object/field/part of the date caused the failure and what was the value. 

To do:
* extend the DataImportReport,
* handle parsing workflow, so errors will be persisted in those reports

bognasoldev89:
Confirmed that extension exists and works

PR #548, opened 2019-07-11
#147: Extend DataImportReport to save parsing errors

jkondrat:


src/main/resources/config/liquibase/changelog/20190709161437_added_entity_OrganizationError.xml
```diff
@@ -0,0 +1,79 @@
...
+
+    <!--
+        Load sample data generated with Faker.js
+        - This data can be easily edited using a CSV editor (or even MS Excel) and
+          is located in the 'src/main/resources/config/liquibase/data' directory
+        - By default this data is applied when running with the JHipster 'dev' profile.
+          This can be customized by adding or removing 'faker' in the 'spring.liquibase.contexts'
+          Spring Boot configuration key.
+    -->
+    <changeSet id="20190709161437-1-data" author="jhipster" context="faker">
```
awalkowiak: I think this is not required. This is added by newer version of Jhipster. We do not need this. I think you can remove this along with organization_error.csv.

jkondrat: done

Issue #542, opened 2019-07-08
Recommendations for improving matching algorithm


johnhbenetech:
Would love someone to do a deep dive on the matching algorithm and come up with recommendations especially after any work with improving geocoding and service taxonomy.

NeilMBenetech:
Duplicate to https://github.com/benetech/ServiceNet/issues/323 ? This has a little more detail, noting prerequisite tasks, so maybe delete the other one ?

johnhbenetech:
@NeilMBenetech yep i closed the other, thanks

jkondrat:
Hi @johnhbenetech : have you noticed any weird matches on staging since the old ones were removed? Do you think the algorithm could still be improved? 
cc: @kkaczmarczyk

PR #588, opened 2019-07-23
#542: Improve the performance of matching

jkondrat:


src/main/java/org/benetech/servicenet/service/dto/LocationDTO.java
```diff
@@ -34,10 +37,88 @@
...
+    @Setter
     private UUID organizationId;
 
+    @Getter
+    @Setter
     private String organizationName;
 
+    private List<GeocodingResultDTO> geocodingResults = new ArrayList<>();
+
+    public String getName() {
```
pmuchowski: you should not mix ordinary getters/setters with lombock

Issue #340, opened 2019-03-19
Search for Services and Locations, not just Organizations


CloVasco:
enable to search by 
-programs and services

NeilMBenetech:
Also by taxonomy / category names - possibly more useful as they are trying to discover new services they don't currently have.

NeilMBenetech:
@johnhbenetech For improved clarity on this item, the user should be able to filter by any of the three major record types : Organization, Service and Location. I propose that every search should target exactly one of these, hence the UI element should be either a set of 3 radio buttons or a drop-down list. The default should be Organization and then any selection made by the user should be persisted and used for all subsequent searches (until the user changes it to something else).

johnhbenetech:
@NeilMBenetech 

Which fields are you talking about specifically for services? There are a lot of fields, so let me know which you feel should be covered in the search scope. 

The geography filtering already allows you to segment by city, zip, county. Are you talking about allowing them to search by street address as well?

NeilMBenetech:
@johnhbenetech For now we should be consistent with the existing search, which is on Name and Alternate Name (correct?) of the targeted record type. 

When we implement https://github.com/benetech/ServiceNet/issues/437 though, the intention there is that any text field in that record type is searchable. We could start though with a subset of fields that we agree are most interesting. But users will surprise us - one random example is wanting to search the Service's Eligibility field for "veteran" as a way to find stuff specific to that type of person.

NeilMBenetech:
@johnhbenetech And yes, street address can also be a great field to search on. "I'm looking for a homeless shelter in San Francisco on Market Street. I don't remember the name." This would use four different filters in combination : taxonomy, geography, record type and targeted search field (last one is this workitem).

johnhbenetech:
@NeilMBenetech 
Ok. Location Name is poorly populated. In many cases we artificially populate it using street address. So lets proceed with name, alt name, and street address for locations.

With regards to UI. Four options (or others if you have a suggestion). A. top level radio (only can search one type), B. top level checkbox (to search multiple types), move either A or B below into filter.

A:
![image](https://user-images.githubusercontent.com/12927553/61809582-380cab00-ae03-11e9-84fa-ddecb075bcc2.png)

B:
![image](https://user-images.githubusercontent.com/12927553/61809880-c6812c80-ae03-11e9-8960-c0d0c2576eba.png)

C:
[...]

PR #593, opened 2019-07-25
#340: Search for Services and Locations, not just Organizations

jkondrat:


src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -98,16 +102,34 @@ public ActivityRecord findOneByOrganizationId(UUID orgId) {
...
+            cb.like(cb.upper(from.get(NAME)), '%' + searchName.trim().toUpperCase() + '%'),
+            cb.like(cb.upper(from.get(ALTERNATE_NAME)), '%' + searchName.trim().toUpperCase() + '%')
+        ));
+    }
+
+    private Predicate searchNames(Predicate predicate, Root<ActivityInfo> root, SearchOn searchOn, String searchName) {
+        if (searchOn.equals(SearchOn.ORGANIZATION)) {
+            return searchNames(predicate, root, searchName);
+        } else if (searchOn.equals(SearchOn.LOCATIONS)) {
+            Join<ActivityInfo, Organization> orgJoin = root.join(ORGANIZATIONS, JoinType.LEFT);
```
pmuchowski: in address filtering you will join the organization again, I think you can just make the join directly in addFilters method

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -98,16 +102,34 @@ public ActivityRecord findOneByOrganizationId(UUID orgId) {
...
+            Join<Organization, Location> locationJoin = orgJoin.join(LOCATIONS, JoinType.LEFT);
```
pmuchowski: the same with the location join

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
jkondrat: done

Issue #555, opened 2019-07-11
Use new geocode structured data in matching algorithm


johnhbenetech:
Once the new structured geocode results are populated, we should switch the matching algorithm to take the coords from that data for all partners.

Another thing to consider along with #542 is whether we should also look at using the structured city, county, zip, address, etc in the matching algorithm. So for instance if we don't have a high confidence coord result do we fall back to the other fields at a lower weighting.

kkaczmarczyk:
I think this should be closed for a while now

PR #595, opened 2019-07-26
#555: Use new geocode structured data in matching algorithm

kkaczmarczyk:


src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
```diff
@@ -33,68 +35,100 @@
...
-        return location1 == null || location1.getPhysicalAddress() == null
-            || location2 == null || location2.getPhysicalAddress() == null;
+    private boolean areProviderCoordEmpty(Location location) {
+        return location.getLongitude() == null || location.getLatitude() == null
+            || location.getLongitude().equals(0.0) || location.getLatitude().equals(0.0);
     }
 
-    private boolean allCoordinatesArePresent(Location location1, Location location2) {
-        return !coordinatesAreMissing(location1) && !coordinatesAreMissing(location2);
+    private boolean areGeocodingResultsNotReliable(List<GeocodingResult> geocodingResults) {
```
jkondrat: I think it would be better to use anyMatch and name this method "areGeocodingResultsReliable"

src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
```diff
@@ -33,68 +35,100 @@
...
         }
 
-        if (allCoordinatesArePresent(location1, location2)) {
-            return countRatioIfCoordinatesArePresent(location1, location2);
-        } else {
-            return countRatioIfCoordinatesAreMissing(location1, location2, context);
+        List<GeocodingResult> baseGeocoding = getGeoCodingResult(location1, context);
+        List<GeocodingResult> partnerGeocoding = getGeoCodingResult(location2, context);
+
+        if (areCoordsNotComparable(baseGeocoding, partnerGeocoding, location1, location2)) {
```
jkondrat: same here, it would be better to have "areCoordsComparable" method

src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
```diff
@@ -33,68 +35,100 @@
...
+        List<GeocodingResult> partnerGeocoding, Location baseLocation, Location partnerLocation) {
+
+        return areGeocodingResultsNotReliable(baseGeocoding) && areProviderCoordEmpty(baseLocation)
+                || areGeocodingResultsNotReliable(partnerGeocoding) && areProviderCoordEmpty(partnerLocation);
     }
 
-    private boolean anyPhysicalAddressIsNull(Location location1, Location location2) {
-        return location1 == null || location1.getPhysicalAddress() == null
-            || location2 == null || location2.getPhysicalAddress() == null;
+    private boolean areProviderCoordEmpty(Location location) {
```
jkondrat: Coords

src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
kkaczmarczyk: Yeah, I had it like that but followed Idea suggestion to inverse the method. I think it may be more readable with anyMatch though.

src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
kkaczmarczyk: fixed

src/main/java/org/benetech/servicenet/matching/counter/LocationSimilarityCounter.java
kkaczmarczyk: fixed

Issue #308, opened 2019-03-16
Ability to dismiss records when done


CloVasco:
Include a way for records to disappear after users:
-import what they need and are done with it
-are not interested in partners' data

So that they can clean their homepage as they work through the list and don't feel overwhelmed by a long list of things that are not relevant to them (any more) 

Questions
-for whom - the user or the team?
-what happens to the record - disappear  from the landing page forever until there's an update from another partner (unless the record is searched)?  is archived somewhere to be retrieved by user? 
-need to create a new button - I am done with this record

CloVasco:
@laurabenetech to add 3 visuals:
Nber 1- visual of link to see  non-matching records that have been dismissed
Nber 2-visual of page when click on Nber 1 to see records that have been dismissed with info (record has been dismissed by person on date. reason for it
Nber 3- feedback window - thanks for your feedback. confirm that you want to dismiss with link to provide reason (non mandatory!!)

laurabenetech:
@CloVasco Here is the Feedback window with note feature. I was able to leverage a design from Exygy. 

![Mismatch confirm_with_note](https://user-images.githubusercontent.com/48697326/55647548-cd235f80-5792-11e9-9595-4df409f8a94c.png)


I'll attach the other visuals before our mtg today.

CloVasco:
@laurabenetech I think we have talked a lot about it and we heard from pilot partners.
Please submit visuals and details for what this feature would look like, including:
-the I am done button - where, name, etc
-and the page where users would be able to see records again after they have been dismissed

laurabenetech:
@CloVasco 
Done button design for Home page:

When user clicks Close (X icon), on the on side-by-side page, the record is dismissed. Hovering over the X with the mouse displays the tool tip: “Done with this record"

Note: For this release, this choice affects the organization level (not the single user level). For a future release, users can choose in Settings to have this be at the individual user or org level. 


<img width="635" alt="Home page Done" src="https://user-images.githubusercontent.com/48697326/57403041-92e02000-718d-11e9-8c5e-ec30c1c1562d.png">

laurabenetech:
@CloVasco 
Done button for Side by side page:
When user clicks Close (X icon), on the on side-by-side page, the record is dismissed. Hovering over the X with the mouse displays the tool tip: “Done with this record”

See visual here for specifications on other minor changes.

For this release, this choice affects the organization level (not the single user level). For a future release, users can choose in Settings to have this be at the individual user or org level. 

[...]

PR #609, opened 2019-08-01
#308: Add ability to hide records on home page in multiple record view

awalkowiak:


src/main/java/org/benetech/servicenet/domain/OrganizationMatch.java
```diff
@@ -39,6 +39,15 @@
     @Column(name = "dismiss_date")
     private ZonedDateTime dismissDate;
 
+    @Column(name = "hidden")
+    private Boolean hidden = false;
+
+    @ManyToOne
+    private User hiddenBy;
```
pmuchowski: I think you can specify the join column

src/main/java/org/benetech/servicenet/domain/view/ActivityInfo.java
```diff
@@ -67,4 +67,8 @@
     public Organization getOrganization() {
         return organizations.iterator().next();
     }
+
+    public Set<OrganizationMatch> getOrganizationMatches() {
```
pmuchowski: there is lombok getter and setter used in this class

src/main/java/org/benetech/servicenet/service/OrganizationMatchService.java
```diff
@@ -58,4 +66,10 @@
     void dismissOrganizationMatch(UUID id, DismissMatchDTO dismissMatchDTO);
 
     void revertDismissOrganizationMatch(UUID id);
+
+    void hideOrganizationMatch(UUID id);
+
+    void hideOrganizationMatch(List<UUID> ids);
```
pmuchowski: I think this should be hideOrganizationMatches

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -211,9 +209,10 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
         }
 
         predicate = addDateFilter(filtersActivityDTO, predicate, root);
+        Join<ActivityInfo, OrganizationMatch> matchJoin = root.join(ORGANIZATION_MATCHES, JoinType.LEFT);
+        predicate = cb.and(predicate, cb.equal(matchJoin.get(HIDDEN), filtersActivityDTO.getHiddenFilter()));
```
pmuchowski: is this filter always not null?

src/main/java/org/benetech/servicenet/service/dto/OrganizationMatchDTO.java
```diff
@@ -25,7 +25,13 @@
 
     private String dismissedByName;
 
-    private ZonedDateTime dismissDate;
```
pmuchowski: why did you removed the dismissDate?

src/main/webapp/app/modules/home/activity-element.tsx
```diff
@@ -1,18 +1,34 @@
...
+import HideRecordButton from 'app/shared/layout/hide-record-button';
+import { toast } from 'react-toastify';
 
 const ActivityElement = props => {
   const maxConflicts = 3;
   const conflictsToDisplay = props.activity.conflicts.slice(0, maxConflicts);
   const areAllDisplayed = props.activity.conflicts.length <= maxConflicts;
 
+  const hideActivity = event => {
+    const matchIds = props.activity.organizationMatches;
```
pmuchowski: you should add the new field to the activity model

src/main/java/org/benetech/servicenet/domain/OrganizationMatch.java
awalkowiak: What's the point of it? You didn't do this for dismissedBy.

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
awalkowiak: I did this with default hidden as false. Do you think it would be better not to have default?

src/main/java/org/benetech/servicenet/service/dto/OrganizationMatchDTO.java
awalkowiak: fixed

src/main/java/org/benetech/servicenet/service/OrganizationMatchService.java
awalkowiak: fixed

src/main/java/org/benetech/servicenet/domain/view/ActivityInfo.java
awalkowiak: fixed

src/main/webapp/app/modules/home/activity-element.tsx
awalkowiak: fixed

src/main/java/org/benetech/servicenet/domain/OrganizationMatch.java
pmuchowski: it's the same like specifying the name of a column, in the code I see some of the relations have the join column and some don't have it
you can just leave it like it is

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
pmuchowski: if there is no possibility to set it to null on the frontend it should be fine

Issue #554, opened 2019-07-11
Use new geocode structured data in geofilters


johnhbenetech:
Once the geocode work is completed we should switch the city, county, zip to use the structured results to provide more normalization and reduce typos, etc.

kkaczmarczyk:
@johnhbenetech a few questions: 

1. should the geocoding filters also hit the partial matches (and/or not ROOFTOP accurate) geocoding results? this depends on how 'strictly' we would like to filter results, I think you may have a better idea about users' perspective.
2. if not - should geo filters hit the provider-given data? (as it is now, but likely with the addition of whitespace trimming of DB results)
3. I assume we should only display the geo filter options that actually exist in geocoding database table - ie. for example there won't be zip codes filling suggestions in filters inputs on UI if there are no geocoding db results containing it - correct?

johnhbenetech:
1. Yes also partial. So like if only city is detected it would still find that

2. Keep it geocode table only

3. Yes only those that exist in table

kkaczmarczyk:
@johnhbenetech Geocoding results are only fetched for Physical Addresses. Nevertheless, Should the geo-filters also run against Postal Address?

johnhbenetech:
Let's stick with just physical

kkaczmarczyk:
I've did an additional quick test on old staging. 
Previously: cities filters were taken from Physical address parsed from Provider's data (screenshot 1+2)
![image](https://user-images.githubusercontent.com/3207119/63028334-eb226e80-beae-11e9-938e-d796b2901cdd.png)
![image](https://user-images.githubusercontent.com/3207119/63028340-ed84c880-beae-11e9-9f99-00489f0e1cb0.png)

Now: cities filters are taken from Geocoding Data

PR #610, opened 2019-08-01
#554: Use new geocode structured data in geofilters

kkaczmarczyk:


src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -122,28 +121,24 @@ private Predicate searchNames(Predicate predicate, From from, String searchName)
 
     private Predicate addLocationFilters(FiltersActivityDTO filtersActivityDTO, Predicate predicate,
         Join<Organization, Location> locationJoin) {
-        Join<Location, PostalAddress> postalAddressJoin = locationJoin.join(POSTAL_ADDRESS, JoinType.LEFT);
-        Join<Location, PhysicalAddress> physicalAddressJoin = locationJoin.join(PHYSICAL_ADDRESS, JoinType.LEFT);
+        Join<Location, GeocodingResult> geocodingResultJoin = locationJoin.join(GEOCODING_RESULTS, JoinType.LEFT);
 
         Predicate updatedPredicate = predicate;
         if (CollectionUtils.isNotEmpty(filtersActivityDTO.getCitiesFilterList())) {
             updatedPredicate = cb.and(updatedPredicate, cb.or(
```
pmuchowski: you don't need the 'or', because it's only one filter here

src/main/java/org/benetech/servicenet/repository/GeocodingResultRepository.java
```diff
@@ -37,4 +39,23 @@
     @Override
     @CacheEvict(value = "addressCache", allEntries = true)
     void deleteAllInBatch();
+
+    @Query("SELECT DISTINCT gr.postalCode FROM GeocodingResult gr "
+        + "LEFT JOIN Location l ON gr in elements(l.geocodingResults) "
```
pmuchowski: why it's done like that and not select form location and join geocoding? is it have better performance?

src/main/java/org/benetech/servicenet/service/impl/ActivityFilterServiceImpl.java
```diff
@@ -45,37 +36,42 @@ public ActivityFilterServiceImpl(PostalAddressRepository postalAddressRepository
...
-        return Stream
-            .of(postalAddressRepository.getDistinctPostalCodesForSystemAccount(systemAccountId),
-                physicalAddressRepository.getDistinctPostalCodesForSystemAccount(systemAccountId))
-            .flatMap(Collection::stream)
-            .collect(Collectors.toCollection(TreeSet::new));
+        if (accountOpt.isEmpty()) {
+            return Collections.emptySet();
+        }
+
+        String systemAccountName = accountOpt.map(SystemAccount::getName).orElse(null);
```
pmuchowski: you can just use accountOpt.get().getName()

PR #611, opened 2019-08-02
#542: Add tests to ensure that the bad matches no longer match

jkondrat:
Organization emails and phones were replaced with fake ones


src/main/java/org/benetech/servicenet/converter/CSVFileConverter.java
```diff
@@ -20,10 +21,9 @@ public CSVFileConverter(String delimiter) {
         }
     }
 
-    @Override
-    public String convert(MultipartFile file) throws IOException {
+    public String convert(InputStream is) throws IOException {
```
pmuchowski: I think it should be private

src/test/java/org/benetech/servicenet/matching/MatchingAccuracyIntTest.java
```diff
@@ -0,0 +1,186 @@
...
+            DocumentUpload upload = new DocumentUpload();
+            upload.setFilename(fileName);
+            uploads.add(upload);
+            data.add(loadCsv(SMC_DIR + fileName + CSV));
+        }
+        MultipleImportData importData = new MultipleImportData(data, uploads, new DataImportReport(), SMC_CONNECT_PROVIDER, true, null);
+        smcConnectDataAdapter.importData(importData);
+    }
+
+    private String loadCsv(String fileName) throws IOException {
```
pmuchowski: maybe you could move it to the utils class

src/main/java/org/benetech/servicenet/converter/CSVFileConverter.java
jkondrat: it's used in integration test

src/test/java/org/benetech/servicenet/matching/MatchingAccuracyIntTest.java
jkondrat: done

Issue #440, opened 2019-05-24
Discovery feature - records not in your own database


CloVasco:
-Landing Page
Enable all users to be able to see all records. We need to enable users to make a clear difference between records that they have and records they don’t have in their directory. Proposition – have 2 tabs on the landing page – one for records they have (what we have today) and one for other agencies they don’t have. Users need to be able to sort/filter/search on those 2 tabs
Laura –we need a visual for those 2 tabs

-Cards
In the discovery tab, what will the cards display on the left as the name of the card if 3 partners have this record with 3 different name..? the easiest solution seems to have 3 different cards for the 3 different partners’ records. 
We’ll need a visual for this as well – or at least a clear understand of what info we put in there. If we use this imperfect but easy solution, we’d have (name of the record on the left – name of the partner and last day of update on the right. That’s it. 
Laura – do you agree? Do you have a better solution?

-Single view 
But if we apply this solution, I think it would be better if users can view those 3 records and “compare them” – meaning if they click on one of those 3 cards, they’d have access to our single view page today and on this page, they would see that 3 partners have it and they’d be able to toggle between them (basically the right potion of our side by side view we have today)

-Side by side view
An idea that would be great if possible would be to provide them a side by side view through which they’d be able to compare those records from partners.. it’d be basically the right potion of our current side by side view both on the right and on the left. John – you didn’t seem to think that this was easy. If it’s not, let’s not do it but at least provide a single view on which they can toggle between records. 

@laurabenetech - please send me through email your comments/suggestions/questions and I’ll answer them through email. Then John will tell you if we actually do this or not. 
@johnhbenetech – Laura told me a while ago that she is not working between tomorrow afternoon and Wednesday. So she’d be able to help next Thursday and Friday.

laurabenetech:
Thanks @CloVasco I'll answer these in separate comments so this doesn't get overwhelming.

laurabenetech:
-Landing Page
Enable all users to be able to see all records. We need to enable users to make a clear difference between records that they have and records they don’t have in their directory. Proposition – have 2 tabs on the landing page – one for records they have (what we have today) and one for other agencies they don’t have. Users need to be able to sort/filter/search on those 2 tabs
Laura –we need a visual for those 2 tabs
@CloVasco The two tab design ok for now since this is a quick solution and you said it’s the simplest to implement. However, I recommend revisiting this for the major overhaul of the design, especially after we get feedback from users about this two-tab design. 
[...]

PR #630, opened 2019-08-14
#440: Make it possible to display records owned by another system acc…

jkondrat:
…ount

### DIFF
--- a/src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
+++ b/src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
@@ -10,9 +10,11 @@ import java.util.UUID;
 import javax.persistence.EntityManager;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
+import javax.persistence.criteria.Expression;
 import javax.persistence.criteria.From;
 import javax.persistence.criteria.Join;
 import javax.persistence.criteria.JoinType;
+import javax.persistence.criteria.Order;
 import javax.persistence.criteria.Predicate;
 import javax.persistence.criteria.Root;
 import org.apache.commons.collections4.CollectionUtils;
@@ -44,6 +46,8 @@ public class ActivityRepository {
 
     private static final String ACCOUNT_ID = "accountId";
     private static final String ID = "id";
+
+    private static final String LAST_UPDATED = "lastUpdated";
 
     private static final String ORGANIZATIONS = "organizations";
     private static final String NAME = "name";
@@ -89,22 +93,30 @@ public class ActivityRepository {
     }
 
     public Page<ActivityInfo> findAllWithFilters(UUID ownerId, String searchName, SearchOn searchOn,
-        FiltersActivityDTO filtersActivityDTO, Pageable pageable) {
+        FiltersActivityDTO filtersActivityDTO, Pageable pageable, Boolean showPartner) {
 
         CriteriaQuery<ActivityInfo> queryCriteria = cb.createQuery(ActivityInfo.class);
         Root<ActivityInfo> selectRoot = queryCriteria.from(ActivityInfo.class);
 
-        queryCriteria.select(selectRoot).distinct(true);
+        queryCriteria.select(selectRoot);
 
-        addFilters(queryCriteria, selectRoot, ownerId, searchName, searchOn, filtersActivityDTO);
-        addSorting(queryCriteria, pageable.getSort(), selectRoot);
+        addFilters(queryCriteria, selectRoot, ownerId, searchName, searchOn,
+            filtersActivityDTO, showPartner);
+        Expression accountMatchExpression = cb.selectCase()
+            .when(cb.equal(selectRoot.get(ACCOUNT_ID), ownerId), 1)
+            .otherwise(0);
+        queryCriteria.groupBy(selectRoot.get(ID), selectRoot.get(ACCOUNT_ID), selectRoot.get(ALTERNATE_NAME),
+            selectRoot.get(NAME), selectRoot.get(RECENT), selectRoot.get(RECOMMENDED),
+            selectRoot.get(LAST_UPDATED), accountMatchExpression);
+
+        addSorting(queryCriteria, pageable.getSort(), selectRoot, accountMatchExpression);
 
         CriteriaQuery<Long> countCriteria = cb.createQuery(Long.class);
         Root<ActivityInfo> selectRootCount = countCriteria.from(ActivityInfo.class);
 
         countCriteria.select(cb.countDistinct(selectRootCount));
 
-        addFilters(countCriteria, selectRootCount, ownerId, searchName, searchOn, filtersActivityDTO);
+        addFilters(countCriteria, selectRootCount, ownerId, searchName, searchOn, filtersActivityDTO, showPartner);
 
         List<ActivityInfo> results = em.createQuery(queryCriteria)
             .setFirstResult((int) pageable.getOffset())
@@ -191,16 +203,19 @@ public class ActivityRepository {
 
     @SuppressWarnings("checkstyle:cyclomaticComplexity")
     private <T> void addFilters(CriteriaQuery<T> query, Root<ActivityInfo> root, UUID ownerId,
-        String searchName, SearchOn searchOn, FiltersActivityDTO filtersActivityDTO) {
+        String searchName, SearchOn searchOn, FiltersActivityDTO filtersActivityDTO, Boolean partnerRecords) {
 
-        Predicate predicate = cb.equal(root.get(ACCOUNT_ID), ownerId);
+        Predicate predicate = cb.conjunction();
+        Predicate isCurrentAccount = cb.equal(root.get(ACCOUNT_ID), ownerId);
+        if (!partnerRecords) {
+            predicate = isCurrentAccount;
+        }
 
-        Join<ActivityInfo, Organization> orgJoin = null;
+        Join<ActivityInfo, Organization> orgJoin = root.join(ORGANIZATIONS, JoinType.LEFT);
         Join<Organization, Location> locationJoin = null;
         Join<Organization, Service> serviceJoin = null;
 
         if (StringUtils.isNotBlank(searchName)) {
-            orgJoin = root.join(ORGANIZATIONS, JoinType.LEFT);
             if (searchOn.equals(SearchOn.ORGANIZATION)) {
                 predicate = searchFields(predicate, orgJoin, searchName, filtersActivityDTO.getSearchFields());
             } else if (searchOn.equals(SearchOn.LOCATIONS)) {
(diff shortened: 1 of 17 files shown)


src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -54,26 +54,40 @@ public ActivityServiceImpl(ActivityRepository activityRepository, RecordsService
...
+            totalElements = partnerActivitiesInfo.getTotalElements();
+            for (ActivityInfo info : partnerActivitiesInfo) {
+                try {
+                    activities.add(getEntityActivity(info, exclusionsMap));
+                } catch (ActivityCreationException ex) {
+                    log.error(ex.getMessage());
+                }
+            }
+        } else {
+            Page<ActivityInfo> activitiesInfo = findAllActivitiesInfoWithOwnerId(systemAccountId, search, searchOn, pageable,
```
pmuchowski: only this part is different, so you can extract the rest of this code outside the if

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -96,15 +96,43 @@ public ActivityRecord findOneByOrganizationId(UUID orgId) {
...
+            .setFirstResult((int) pageable.getOffset())
+            .setMaxResults(pageable.getPageSize())
+            .getResultList();
+
+        Long total = em.createQuery(countCriteria).getSingleResult();
+
+        return new PageImpl<>(results, pageable, total.intValue());
+    }
+
+    public Page<ActivityInfo> findPartnerRecordsWithFilters(UUID ownerId, String searchName, SearchOn searchOn,
```
pmuchowski: what is the difference between those two methods? is it only the parameter passed to the addFilters?
is the separate method necessary in this case?

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -216,11 +248,20 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
...
-        if (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList())) {
+        boolean hasPartnerFilters = (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList()));
+        if (hasPartnerFilters || partnerRecords) {
             Join<OrganizationMatch, Organization> matchOrgJoin = matchJoin.join(PARTNER_VERSION, JoinType.LEFT);
-            Join<Organization, SystemAccount> accountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
+            Join<Organization, SystemAccount> matchAccountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
 
-            predicate = cb.and(predicate, accountJoin.get(ID).in(filtersActivityDTO.getPartnerFilterList()));
+            if (partnerRecords) {
+                predicate = cb.and(predicate, cb.notEqual(matchAccountJoin.get(ID), ownerId));
```
pmuchowski: if there are any partner filters I think it should be enough to make sure the ownerId is not in the filtersActivityDTO.getPartnerFilterList and if there are none you can add this predicate

jkondrat: I don't think it would work, in that case partnerFilterList filters on matches, not the selected organization

pmuchowski: OK, I see now

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
jkondrat: done

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
jkondrat: done

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -242,24 +283,27 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
...
         }
 
-        query.where(predicate);
+        return predicate;
     }
 
-    private void addSorting(CriteriaQuery<ActivityInfo> queryCriteria, Sort sort, Root<ActivityInfo> root) {
-        setOrder(queryCriteria, root, sort, RECENT);
-        setOrder(queryCriteria, root, sort, RECOMMENDED);
+    private void addSorting(CriteriaQuery<ActivityInfo> queryCriteria, UUID ownerId, Sort sort,
```
pmuchowski: the ownerId is not used anywhere

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -190,17 +218,21 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
...
-    private <T> void addFilters(CriteriaQuery<T> query, Root<ActivityInfo> root, UUID ownerId,
-        String searchName, SearchOn searchOn, FiltersActivityDTO filtersActivityDTO) {
-
-        Predicate predicate = cb.equal(root.get(ACCOUNT_ID), ownerId);
+    private <T> Predicate addFilters(CriteriaQuery<T> query, Root<ActivityInfo> root, UUID ownerId,
+        String searchName, SearchOn searchOn, FiltersActivityDTO filtersActivityDTO, Boolean partnerRecords) {
+
+        Predicate predicate;
+        if (partnerRecords) {
+            predicate = cb.notEqual(root.get(ACCOUNT_ID), ownerId);
```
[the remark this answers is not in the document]
jkondrat: why is that? when looking for partner records, we don't want main organization to be from the current account

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -216,11 +248,20 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
...
 
-        if (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList())) {
+        boolean hasPartnerFilters = (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList()));
+        if (hasPartnerFilters || partnerRecords) {
             Join<OrganizationMatch, Organization> matchOrgJoin = matchJoin.join(PARTNER_VERSION, JoinType.LEFT);
-            Join<Organization, SystemAccount> accountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
+            Join<Organization, SystemAccount> matchAccountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
 
-            predicate = cb.and(predicate, accountJoin.get(ID).in(filtersActivityDTO.getPartnerFilterList()));
+            if (partnerRecords) {
```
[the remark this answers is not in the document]
jkondrat: I thought that we only want to filter on oganization, not it's matches when the partner records are on

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -216,11 +248,20 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
         Join<ActivityInfo, OrganizationMatch> matchJoin = root.join(ORGANIZATION_MATCHES, JoinType.LEFT);
         predicate = cb.and(predicate, cb.equal(matchJoin.get(HIDDEN), filtersActivityDTO.getHiddenFilter()));
 
-        if (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList())) {
+        boolean hasPartnerFilters = (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList()));
+        if (hasPartnerFilters || partnerRecords) {
```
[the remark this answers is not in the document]
jkondrat: We will then have to repeat joins in two if branches

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -89,22 +93,46 @@ public ActivityRecord findOneByOrganizationId(UUID orgId) {
...
         CriteriaQuery<Long> countCriteria = cb.createQuery(Long.class);
         Root<ActivityInfo> selectRootCount = countCriteria.from(ActivityInfo.class);
 
-        countCriteria.select(cb.countDistinct(selectRootCount));
+        countCriteria.select(cb.count(selectRootCount));
 
-        addFilters(countCriteria, selectRootCount, ownerId, searchName, searchOn, filtersActivityDTO);
+        predicate = addFilters(countCriteria, selectRootCount, ownerId, searchName, searchOn, filtersActivityDTO, false);
+        if (showPartner) {
+            countCriteria.where(cb.or(
```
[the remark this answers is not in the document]
jkondrat: done

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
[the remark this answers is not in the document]
jkondrat: done

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
jkondrat: done

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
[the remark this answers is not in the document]
pmuchowski: but you're not looking only for partner records, right?
in case the partner records is true you're calling this method twice which will duplicate the name and location search, you can simplify this

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -216,11 +230,23 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
...
-            Join<Organization, SystemAccount> accountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
-
-            predicate = cb.and(predicate, accountJoin.get(ID).in(filtersActivityDTO.getPartnerFilterList()));
+            Join<Organization, SystemAccount> matchAccountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
+            if (partnerRecords) {
+                predicate = cb.and(predicate, cb.or(cb.equal(root.get(ACCOUNT_ID), ownerId),
+                    cb.notEqual(matchAccountJoin.get(ID), ownerId)));
+                if (hasPartnerFilters) {
+                    Join<Organization, SystemAccount> accountJoin = orgJoin.join(ACCOUNT);
+                    predicate = cb.and(predicate, cb.or(cb.equal(root.get(ACCOUNT_ID), ownerId),
```
pmuchowski: this filter will not work, it will display all organizations of current user, it will not filter them by matched account
remove the `cb.or(cb.equal(root.get(ACCOUNT_ID), ownerId)`

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -191,16 +203,18 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
 
     @SuppressWarnings("checkstyle:cyclomaticComplexity")
     private <T> void addFilters(CriteriaQuery<T> query, Root<ActivityInfo> root, UUID ownerId,
-        String searchName, SearchOn searchOn, FiltersActivityDTO filtersActivityDTO) {
+        String searchName, SearchOn searchOn, FiltersActivityDTO filtersActivityDTO, Boolean partnerRecords) {
 
-        Predicate predicate = cb.equal(root.get(ACCOUNT_ID), ownerId);
+        Predicate predicate = root.isNotNull();
```
pmuchowski: use cb.conjunction() instead

src/main/java/org/benetech/servicenet/repository/ActivityRepository.java
```diff
@@ -216,11 +230,23 @@ private Predicate addDateFilter(FiltersActivityDTO filtersActivityDTO, Predicate
...
-        if (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList())) {
+        boolean hasPartnerFilters = (CollectionUtils.isNotEmpty(filtersActivityDTO.getPartnerFilterList()));
+        if (partnerRecords || hasPartnerFilters) {
             Join<OrganizationMatch, Organization> matchOrgJoin = matchJoin.join(PARTNER_VERSION, JoinType.LEFT);
-            Join<Organization, SystemAccount> accountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
-
-            predicate = cb.and(predicate, accountJoin.get(ID).in(filtersActivityDTO.getPartnerFilterList()));
+            Join<Organization, SystemAccount> matchAccountJoin = matchOrgJoin.join(ACCOUNT, JoinType.LEFT);
+            if (partnerRecords) {
+                predicate = cb.and(predicate, cb.or(cb.equal(root.get(ACCOUNT_ID), ownerId),
```
pmuchowski: I think you can remove `cb.equal(root.get(ACCOUNT_ID), ownerId)`, because organization cannot have a match on organization with the same account, so `cb.notEqual(matchAccountJoin.get(ID), ownerId)` will be always true for current account organizations

Issue #602, opened 2019-07-30
Store scheduler config in DB to prevent auto schedule on every deploy


johnhbenetech:

johnhbenetech:
@kkaczmarczyk As part of this task can you confirm one thing.

if i disable a job I would expect it to not run at all until I manually fire it again. I've noticed sometimes where I know i disabled it, but then it seems to run a week later. It's possible it was redeployed in the inbetween, but i think there was a few times in prod where i know i didnt redeploy. I could be mistaken though :)

kkaczmarczyk:
@johnhbenetech  there shouldn't be any issues with disabled tasks now

PR #633, opened 2019-08-17
#602: Store scheduler config in DB to prevent auto schedule on every …

kkaczmarczyk:
…deploy


src/main/java/org/benetech/servicenet/config/QuartzConfig.java
```diff
@@ -47,35 +43,22 @@
...
     public SchedulerFactoryBean schedulerFactoryBean() {
-        SchedulerFactoryBean scheduler = new SchedulerFactoryBean();
+        SchedulerFactoryBean schedulerFactoryBean = new SchedulerFactoryBean();
 
         AutowiringBeanJobFactory jobFactory = new AutowiringBeanJobFactory();
         jobFactory.setApplicationContext(applicationContext);
-        scheduler.setJobFactory(jobFactory);
-
-        List<BaseJob> jobs = new ArrayList<>();
-        jobs.add(edenDataUpdateJob);
```
pmuchowski: you can remove the autowired jobs if they are not used anymore.

src/main/java/org/benetech/servicenet/service/impl/SchedulerServiceImpl.java
```diff
@@ -52,6 +54,31 @@
...
+        jobs.add(uwbaTaxonomyUpdateJob);
+
+        jobs
+            .forEach(job -> {
+                try {
+                    if (scheduler.getTrigger(job.getInitTrigger().getKey()) == null) {
+                        scheduler.scheduleJob(edenDataUpdateJob.getJobDetail(), edenDataUpdateJob.getInitTrigger());
+                    }
+                } catch (SchedulerException e) {
+                    log.error(e.getMessage());
```
pmuchowski: I think you should log the entire exception `log.error(e.getMessage(), e);`

src/main/resources/config/liquibase/tables_postgres.sql
```diff
@@ -0,0 +1,187 @@
+-- Thanks to Patrick Lightbody for submitting this...
+--
+-- In your Quartz properties file, you'll need to set 
+-- org.quartz.jobStore.driverDelegateClass = org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
```
pmuchowski: where is this file used?

src/main/java/org/benetech/servicenet/config/QuartzConfig.java
kkaczmarczyk: fixed

src/main/java/org/benetech/servicenet/service/impl/SchedulerServiceImpl.java
kkaczmarczyk: fixed

src/main/resources/config/liquibase/tables_postgres.sql
kkaczmarczyk: it's meant for a re-creation of Quartz DB tables - I have added info to the README.

Issue #585, opened 2019-07-23
Search - retain user filter and sorting settings across searches and sessions


NeilMBenetech:
Whenever a user changes any of the filtering and sorting options on the main Search page, persist those settings and automatically apply them when they return to the search page. (Do not retain the free text in the search box though, that should be newly added each time the initiate a search). 

Secondarily, allow the user to save different named sets of settings so they could quickly choose one from their list and have those settings applied. Example, one setting called "Food Pantries in Santa Clara County" might have geography filter set and category/taxonomy filter set to those values. Another one might be "Organizations updated in past 10 days" with corresponding record type and filter options.

This depends upon these various filter and sorting options to be coded and available in Service Net, of course, but all of them don't need to be coded to begin this capability. Once new filters are later added, they can be incorporated in this Save Settings functionality.

johnhbenetech:
@NeilMBenetech Do you have a UI suggestions for the 'Save Search' and Search Select features?

johnhbenetech:
In the mean time we can go ahead with persisting the filters between record views,

NeilMBenetech:
@johnhbenetech It probably warrants its own accordion named "Saved searches". Then upon expansion the user has a text box where they can type a new name (must be unique) and click a Save button. Below that is the list of that particular user's previously saved searches. When they click on any of them it applies it's filter settings and does the search automatically.

bognasoldev89:
I checked this new feature and it works just fine.
Below some simple examples of retaining filters and saving filters:

![retain filters log out working](https://user-images.githubusercontent.com/51324797/63854228-fdb3a200-c99c-11e9-9b79-758d94a0ea55.gif)

![saving filters working](https://user-images.githubusercontent.com/51324797/63854109-a7defa00-c99c-11e9-9e45-0329161243c3.gif)

NeilMBenetech:
It appears to be working for me too, thank you. I did find a some other issues but will open new tasks for those.

bognasoldev89:
Unfortunately I have to reopen this issue. I found that user can't save filter with "Partner Sources" filter set up.

bognasoldev89:
User can save filter with "Partner Sources" now - bug is fixed and the issue can be closed.

![585 partner sources working](https://user-images.githubusercontent.com/51324797/64115814-e0ae1300-cd90-11e9-8c7f-cbf209e36b7d.gif)

PR #673, opened 2019-08-30
#585: Fix saving filters with Partner Sources

pmuchowski:


src/main/webapp/app/modules/home/filter-activity.tsx
```diff
@@ -203,6 +204,17 @@ export class FilterActivity extends React.Component<IFilterActivityProps, IFilte
     this.props.updateActivityFilter({ ...this.props.activityFilter, showPartner: !onlyShowMatching });
   };
 
+  getPartnerListValues = () => {
+    const partnerList = [];
+
+    _.forEach(this.props.selectedPartner, partnerId => {
```
awalkowiak: I think this.props.selectedPartner could have better naming if this contains list with partnerFilterList

Issue #632, opened 2019-08-17
Update OrganizationMatches when Org has changed (after upload)


kkaczmarczyk:
It would make sense to update existing Organization Matches in case if Org has changed, since the previous upload.

bognasoldev89:
I managed to "unmatch" and "rematch" organizations and also change them - all of these actions had an impact on OrganizationMatches ("unmatching" took a while though).

Here some examples of Organization Match before "unmatching", rematched and changed (ID of matches is different and Timestamp is changed):

![632 changed org name](https://user-images.githubusercontent.com/51324797/65243349-7911fa80-dae8-11e9-99dc-5df5e1dcf5eb.png)

![632 changed org name 2](https://user-images.githubusercontent.com/51324797/65243351-79aa9100-dae8-11e9-8e1a-176962a9f241.png)

![632 changed org name 3](https://user-images.githubusercontent.com/51324797/65243352-79aa9100-dae8-11e9-8d50-75a1f053db62.png)

![632 changed org name 4](https://user-images.githubusercontent.com/51324797/65243353-79aa9100-dae8-11e9-9502-1094b8111024.png)

kkaczmarczyk:
@johnhbenetech would like to have some more documentation regarding this issue and its cases in the Wiki or is that pretty clear?

johnhbenetech:
@kkaczmarczyk I think it would be helpful

PR #677, opened 2019-09-04
#632: Update OrganizationMatches when org has changed

jkondrat:


src/main/java/org/benetech/servicenet/service/impl/OrganizationMatchServiceImpl.java
```diff
@@ -218,15 +240,27 @@ public void createOrUpdateOrganizationMatches(Organization organization, Matchin
                 .map(m -> m.getPartnerVersion().getId())
                 .collect(Collectors.toList());
 
-            List<Organization> notMatchedOrgs = findNotMatchedOrgs(currentMatchesIds, organization.getAccount().getName());
+            List<Organization> partnerOrganizations = findOrganizationsExcept(organization.getAccount().getName());
 
-            findAndPersistMatches(organization, notMatchedOrgs, context);
+            List<OrganizationMatch> matches = findAndPersistMatches(organization, partnerOrganizations, context);
+            removeObsoleteMatches(matches, currentMatches);
```
kkaczmarczyk: would it make more sense to rename currentMatches -> previousMatches? It looks like it is used that way in removeObsoleteMatches method.

jkondrat: they become 'previousMatches' just after findAndPersistMatches is called.

Issue #623, opened 2019-08-08
Come up with proper approach for populating taxonomy filter dropdown


johnhbenetech:
So this is my current understanding of how taxonomy filter dropdown is populated:

* Taxonomies are pulled from various partner systems, either through pulling values from CSV uploads, or by hitting specific taxonomy API 

* The dropdown then shows all taxonomies from all partners

So this can get complicated for a few reasons:

1. There exist taxonomies in the dropdown that have no records associated with them (they are simply available through the API endpoint). This can be confusing when a user repeated runs into issues where no records are returned.

2. As we've discussed previously, the complexity increases with the 'only show matching' off. If you select a taxonomy from a partner system, none of your records will return. If you pick a taxonomy from your system, no partner records will return.

@NeilMBenetech what's the recommended path forward here? Maybe for 1 we simply dont show values that have no records associated with them.

Number 2 is daunting and gets at the need for the rosetta stone concept. How do you think we should progress here?

johnhbenetech:
yes

________________________________

NeilMBenetech:
@johnhbenetech I think the only way to solve this is via the Taxonomy Rosetta Stone. https://github.com/benetech/ServiceNet/issues/319 

It is an important issue and for the reasons you note above, it can cause confusion for our users leading to less trust in the system.

NeilMBenetech:
@johnhbenetech Oh and absolutely, we should not show taxonomy terms in the auto-suggest / filter that do not have any records associated with them. 

It would be ideal if the other filters could also be applied to what gets autosuggested in the taxonomy filter. For example "There are no Food Pantries in Marin County" so don't auto-suggest Food Pantries when Marin County is in the geo filter...and so on. This is probably pretty challenging though so feel free to decline at this time.

bognasoldev89:
At this moment the dropdown list is populated for other providers, not the one that user is logged. I think it needs a little more work.

Here example of how it works right now:
![Peek 2019-10-17 16-44](https://user-images.githubusercontent.com/51324797/67019901-a8268680-f0fd-11e9-92dd-85aa8daab332.gif)

NeilMBenetech:
@bognasoldev89 Yes, it should autosuggest the taxonomy of the current user, when no particular partner data source is selected.

NeilMBenetech:
[...]

PR #696, opened 2019-10-10
#623: Hide the taxonomies that have no records associated with them

jkondrat:


src/main/webapp/app/modules/home/filter-activity.tsx
```diff
@@ -430,12 +458,23 @@ export class FilterActivity extends React.Component<IFilterActivityProps, IFilte
   }
 }
 
+function getTaxonomyOptions(taxonomyMap) {
+  const taxonomyOptions = {};
+  for (const key in taxonomyMap) {
+    if (Object.prototype.hasOwnProperty.call(taxonomyMap, key)) {
```
kkaczmarczyk: can we use lodash _.has here?

src/main/webapp/app/modules/home/filter-activity.tsx
```diff
@@ -430,12 +458,23 @@ export class FilterActivity extends React.Component<IFilterActivityProps, IFilte
   }
 }
 
+function getTaxonomyOptions(taxonomyMap) {
+  const taxonomyOptions = {};
+  for (const key in taxonomyMap) {
```
kkaczmarczyk: Can we use either .forEach or _.forEach here? Whichever is being used mostly in the project?

src/main/java/org/benetech/servicenet/service/impl/ActivityFilterServiceImpl.java
```diff
@@ -67,8 +70,15 @@ public ActivityFilterServiceImpl(GeocodingResultRepository geocodingResultReposi
...
     @Override
-    public Set<String> getTaxonomies() {
-        return taxonomyRepository.findAll().stream().map(Taxonomy::getName).collect(Collectors.toSet());
+    public Map<String, Set<String>> getTaxonomies() {
+        List<Taxonomy> taxonomies = taxonomyRepository.findAssociatedTaxonomies();
+        Map<String, Set<String>> taxonomiesByProvider = new HashMap<>();
+        for (Taxonomy taxonomy : taxonomies) {
+            Set<String> providersTaxonomies = taxonomiesByProvider.getOrDefault(taxonomy.getProviderName(), new HashSet<>());
+            providersTaxonomies.add(taxonomy.getName());
+            taxonomiesByProvider.put(taxonomy.getProviderName(), providersTaxonomies);
```
kkaczmarczyk: If performance is a concern here, this can changed to just adding new elements to the Value of the Map. If not - it is fine.

src/main/webapp/app/modules/home/filter-activity.tsx
jkondrat: we wouldn't be able to return from inside of the loop

Issue #680, opened 2019-09-05
Show more information about last updated


NeilMBenetech:
When viewing your own record, whether or not you are comparing to a partner's "matched" record or not, please change the user interface. For any record that lacks one of these data elements, just put "unknown" as the value for items 1 through 4 below.

1. Organization records : For those data exports that are giving us valid data about the most recent "Complete Review" (in iCarol it is called "LastVerifiedOn and LastVerifiedBy"), please add a line above "Last updated" called "Last complete review " and put the corresponding date, time, person's name and source of information (if available - I don't think iCarol has this last field yet).

2. Organizational records: Rename "Last updated" to be "Last partial update". 

3. Service records and Location records: Please add "Last complete review" and "Last partial update"  to show, formatted the same way as specified in item 1 and 2 above.

4. Service Records: For those exports that have "usage" data (in iCarol it might be there called NumberOfReferrals), please for now add a line under item "Last partial update" that reads: "Number of times used in past 12 months : " and append the usage data. This will likely need some refinement in the future (for example to characterize as "low, medium or high for this data partner" but for now let's keep it simple.

5. Please let me know, for each partners' exports, which data elements above are currently available and which are not.

johnhbenetech:
So @NeilMBenetech looking at the various data we have access to:

----

**LAAC:** Only field I see is a date field called "changed". All of those dates occur in the last 2 months so I'm not sure if this constitutes anything related to 'verification'.

**Recommended Action:** Pull nothing right now

---

**SMC:** No obvious data element present for verification/updates sent in the HSDS 

**Recommended Action:** Pull nothing right now

---

**HealthLeads:** No obvious data element present for verification/updates sent in the HSDS 

**Recommended Action:** Pull nothing right now

---

UWBA/EDEN (iCarol): 
 "lastVerifiedOn": present at Agency, Site, and Program level

@NeilMBenetech I don't see LastVerifiedBy within the API response. Looking through the API docs - is it possible it's only available via CSV export? I also dont see anything related to referral count or usage.

**Recommended Action:** Add and display LastVerifiedOn to Organization, Location, and Service as indicated by Neil above.


---

**ShelterTech**:

* Organization (368 Total)
** "certified" - 344 True, 24 False
** "verified_at" - 217 Blank, 43 in 2019, 99 in 2018, 9 in 2017
** "certified_at" - 221 Blank, 111 in 2019, 36 in 2018

* Services (1397 Total)
** "certified" - 1397 Blank
[...]

PR #717, opened 2019-10-25
#680: Put unknown instead of blank for last updated/verified dates

jkondrat:


src/main/webapp/i18n/en/multiRecordView.json
```diff
@@ -19,6 +19,7 @@
     "from": "From ",
     "lastUpdated": "Last partial update ",
     "lastCompleteReview": "Last complete review ",
+    "unknown": "unknown",
```
kkaczmarczyk: I guess we can make it start with capital letter - "Unknown"

jkondrat: This becomes a part of sentence - Last review unknown
should I still change the letter?

kkaczmarczyk: ok I imagined it differently

Issue #746, opened 2019-11-27
Handle the case when physical address is null to prevent error on organization detail page


johnhbenetech:
see #698

jkondrat:
It turns out that there are 42 records with empty physical_address; I have changed the UI to skip this section for these records, and display postal address in the dropdown instead.
> "Seniors EyeCare Program (SEP)"
> "BART"
> "Crystal Springs Trail County Park"
> "James V. Fitzgerald Marine Reserve"
> "Wunderlich Park"
> "RecycleWorks"
> "Overeaters Anonymous of the Mid-Peninsula"
> "Northern California Gamblers Anonymous"
> "Transgender Care Clinic for Trans Youth"
> "Healthy Homes"
> "Daybreak"
> "Child & Adolescent Hotline & Prevention Program"
> "Crisis Intervention and Suicide Prevention Center"
> "Crisis Hotline"
> "Alcohol and Drug Helpline"
> "Bay Area Legal Aid - Marin Regional Office - Location"
> "Children and Adults with Attention Deficit/Hyperactivity Disorder"
> "Healthy Homes"
> "Daybreak"
> "Child & Adolescent Hotline & Prevention Program"
> "Crisis Intervention and Suicide Prevention Center"
> "Crisis Hotline"
> "Alcohol and Drug Helpline"
> "Community Health., Mobile Clinic"
> "Explorers' Program"
> "Healthy Homes"
> "Daybreak"
> "Child & Adolescent Hotline & Prevention Program"
> "Crisis Intervention and Suicide Prevention Center"
> "Crisis Hotline"
> "Alcohol and Drug Helpline"
> "Healthy Homes"
> "Daybreak"
> "Child & Adolescent Hotline & Prevention Program"
> "Crisis Intervention and Suicide Prevention Center"
> "Crisis Hotline"
> "Alcohol and Drug Helpline"
> "Community Service Center"
> "California Youth Crisis Line"
> "Bay Area Legal Aid - Marin Regional Office - Location"
> "BART"
> "Compassionate Friends"
>

bognasoldev89:
I checked those cases and error does not occur anymore and it seems to be working just fine now. For example:

![746](https://user-images.githubusercontent.com/51324797/70043508-e6e6a200-15c0-11ea-9cde-5833a5eda10b.png)

@johnhbenetech  @NeilMBenetech could you check if that's ok? Here comparison of UI for the modified record and a regular one:

[...]

PR #752, opened 2019-11-29
#746: Handle the case when physical address is null to prevent error …

jkondrat:
…on organization detail page

### DIFF
--- a/src/main/webapp/app/modules/conflicts/shared/components/location/single-location-details.tsx
+++ b/src/main/webapp/app/modules/conflicts/shared/components/location/single-location-details.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Col, Row, Button } from 'reactstrap';
+import { Col, Row } from 'reactstrap';
 import '../../shared-record-view.scss';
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/modules/conflicts/shared/components/location/single-location-details.tsx
```diff
@@ -45,6 +45,14 @@ export class SingleLocationDetails extends React.Component<ISingleLocationDetail
       .replace(/0\./, '')}`
   };
 
+  static getSelectOption = (record, locationNumber) => ({
+    value: locationNumber,
+    label:
+      record.physicalAddress !== null
+        ? `${record.physicalAddress.address1} ${record.physicalAddress.city}`
+        : `${record.postalAddress.address1} ${record.postalAddress.city}`
```
awalkowiak: What in case if both physicaladdress and postaladdress will be null? Is this possible?

jkondrat: > What in case if both physicaladdress and postaladdress will be null? Is this possible?

yes, it's possible. I'll display location name in such case

Issue #793, opened 2020-01-27
Investigate displaying Service name for SMCConnect provider


bognasoldev89:
During January regression I noticed, that some of Service names disappeared from dropdown lists on single and multi record view for records from SMCConnect provider.
I haven't noticed this issue for other providers.

Exemplary records:
https://benetechservicenet.org/#/single-record-view/fbeedabf-d87f-455b-89f8-5a784a0053bb
https://benetechservicenet.org/#/single-record-view/6cc10df8-6262-4065-88e9-7b0868aecc88
https://benetechservicenet.org/#/single-record-view/0252fbf9-26cf-4f1e-9db5-da7e75a64fc2

Gif with captured bug:
![sevices dropdown list bug](https://user-images.githubusercontent.com/51324797/73181810-399f8e80-4118-11ea-92fb-e9eeb912b765.gif)

kdondziak:
@johnhbenetech @NeilMBenetech I investigated the issue and this happen when Service's name is empty. What should be displayed in that case? Maybe some validation should be added (name not empty)?

awalkowiak:
@kdondziak we had previously service name not null restriction, but this this restriction was removed, because we wanted to keep as much data as we could. Perhaps in this case we could use organization name + '- Service X' where X would be some number that could help identify services for viewing purposes. 

Alternatively during data upload for all providers we could add behavior which in case if there will be no name for service, then we save name with the same pattern as in LAAC provider (which is the mentioned org name + service postfix)

cc: @johnhbenetech @NeilMBenetech

NeilMBenetech:
@johnhbenetech  @kdondziak this is what I was thinking too, good idea John: "...we could use organization name + '- Service X' where X would be some number that could help identify services for viewing purposes"

kdondziak:
@johnhbenetech  @NeilMBenetech During slack conversation John agreed that labels should be like 'Service 1, Service 2' without organization name. Please tell me if I am wrong.

johnhbenetech:
Let's go with Service 1, Service 2

bognasoldev89:
In that case, it looks good :).

Example:
![793](https://user-images.githubusercontent.com/51324797/73846017-0a79d300-4824-11ea-924c-ca1b7d8cecf2.gif)

PR #806, opened 2020-01-31
 #793: Fix empty select when service name is empty

kdondziak:


src/main/webapp/app/modules/conflicts/shared/components/service/services-details.tsx
```diff
@@ -29,6 +30,16 @@ export class ServicesDetails extends React.Component<IServicesDetailsProp, IServ
     this.setState({ serviceNumber });
   };
 
+  getServiceDisplayLabel = (s, index) => {
+    if (!!s.service && !_.isEmpty(s.service.name)) {
+      return { value: index, label: s.service.name };
+    }
+    return {
+      value: index,
+      label: `${translate('serviceNetApp.service.detail.title')} ${index}`
```
awalkowiak: Using the index here won't give us certainty that this number will be increased by 1 for each case (we might have labels which could mislead users). I suggest that this value should be incremented by 1 always.

Issue #780, opened 2020-01-10
Allow users to express record matches manually at the Service and Location level


NeilMBenetech:
When a user is comparing their own record to partner records, they might spend considerable time deciding which partners' Service records actually match to their own Service records. Once they have done that work, it should be retained so they don't have to do that again in the future, for those records.

So please enable users to manually mark their own Service to be a match to 0, 1 or many partner Service records and when they are viewing their own Service record, they can quickly navigate to the partner Service records they had previously marked as "matches". 

This "quick navigation" might be a hyperlink listed in their own Service record that, when clicked, brings the corresponding comparison Service record to the right-hand column. The text for the hyperlink should be [Partner Name] + [Matching Service Name] so for example: "Health Leads - Larry's Food Pantry" which should be sufficient for this and not require more information (such as the Service's parent Organization record name).

Any manual match saved by a user should be available to all other users from that partner. Users should also be able to delete these manual matches.

The data generated from these manual matches could also be used by us to help us check and tune our automated Service-level matching laid out here : https://github.com/benetech/ServiceNet/issues/699

johnhbenetech:
@NeilMBenetech we will need a UI suggestion for this. Is that something you'd like to handle?

kdondziak:
@johnhbenetech @NeilMBenetech In the description is mentioned that #699 should be done first. Am I right? Also what about UI suggestion?

NeilMBenetech:
Hi @kdondziak actually I think this one should be done before #699 so that data from this one could feed the algorithm used there. 

@johnhbenetech  As for UI, I think the original write-up of this addresses how to navigate to the manually-matched records, but you're probably looking for a UI suggestion for how to mark the matches, manually, in the first place.

I recommend that when you are viewing your own record's detail in the left column and have a partner's record in comparison in the right column, there is a button below the Service name that reads something like "Matches my Service record". If the shown partner record had already been marked as a match, the button would change to say "Unmatch this record".

The same would apply for Location records as originally specified in this item.

In thinking further about this, I think it makes sense to allow for this matching to happen at the Organization level too. So, all three levels would be able to be indicated as a match, independently. That's because, while the partner's Organization might match my own (and I want to manually confirm that), every child Service and Location may or may not match to my own.

Further, I anticipate some filters on the main Search page (in a later Github task) where users can go directly to just records they had previously designated as Matched.

abaranowska1:
[...]

Issue #801, opened 2020-01-30
Fix matching records at the Location level


bognasoldev89:
Issue #698 doesn't give results it used to anymore.
Migrating database from prod to staging is suspected, but it needs further investigation.

bognasoldev89:
@jkondrat for most organizations this feature work as before:

![801 3](https://user-images.githubusercontent.com/51324797/74021011-0c19d700-499b-11ea-82b2-0581498cab3e.gif)

One of the mentioned in #698 organizations doesn't rotate to Location I would expect though. SAN FRANCISCO OFFICE OF ECONOMIC AND WORKFORCE DEVELOPMENT (OEWD), WORKFORCE DEVELOPMENT DIVISION  from UWBA is matched to  Employment Development Department from healthleads. Both have 801 Turk Street San Francisco Location. But it's not matched:

![801](https://user-images.githubusercontent.com/51324797/74021444-daedd680-499b-11ea-936e-6ae18c87bcd8.gif)

Another thing. It seems like: if a record has 2 or more Locations and its match has 1 Location - Location on the left side is not rotated to match (checked on MISSION NEIGHBORHOOD HEALTH CENTER from UWBA to Mission Neighborhood Health Center from healthleads):

![801 2](https://user-images.githubusercontent.com/51324797/74021588-14bedd00-499c-11ea-80b4-6c84b4fa9ada.gif)

bognasoldev89:
Yes, seems fine now, I didn't find any matches that would rotate badly.

Example from my previous comment:
![801 working again](https://user-images.githubusercontent.com/51324797/74253970-18ca6200-4cf0-11ea-9306-2fe491982522.gif)

PR #849, opened 2020-02-26
#780: Allow users to express record matches manually at the Service and Location level

kdondziak:


src/main/java/org/benetech/servicenet/repository/LocationMatchRepository.java
```diff
@@ -16,7 +18,10 @@
 public interface LocationMatchRepository extends JpaRepository<LocationMatch, UUID> {
     Set<LocationMatch> findByLocationId(UUID locationId);
 
-    Optional<LocationMatch> findByLocationIdAndMatchingLocationId(UUID locationId, UUID matchingLocationId);
+    @Query("SELECT locationMatch FROM LocationMatch locationMatch "
```
jkondrat: is this any different from the jpa-generated method?

src/main/java/org/benetech/servicenet/service/impl/LocationMatchServiceImpl.java
```diff
@@ -35,6 +69,16 @@ public LocationMatch saveOrUpdate(LocationMatch locationMatch) {
 
     @Override
     public void delete(UUID locationId, UUID matchingLocationId) {
-        locationMatchRepository.deleteByLocationIdAndMatchingLocationId(locationId, matchingLocationId);
+        Optional<LocationMatch> existingMatch = locationMatchRepository.findByLocationIdAndMatchingLocationId(
```
jkondrat: Perhaps we should add some constraint to ensure that there can be only one location/service match for given ids

jkondrat: what would happen if we tried saving the same match concurrently?

src/main/webapp/app/modules/conflicts/shared/components/service/single-service-details.tsx
```diff
@@ -84,14 +127,43 @@ export class SingleServiceDetails extends React.Component<ISingleServiceDetailsP
...
+    const unmatchButton = (
+      <Button onClick={this.unmatchService} replace color="danger">
+        <FontAwesomeIcon icon="arrow-left" />{' '}
+        <span className="d-none d-md-inline">
+          <Translate contentKey="multiRecordView.unmatchThisRecord" />
+        </span>
+      </Button>
+    );
+
+    const matchOrUnMatchButton = (
```
jkondrat: rename to 'toggleMatch'

src/main/webapp/app/modules/conflicts/shared/shared-record-view.reducer.ts
```diff
@@ -75,6 +91,16 @@ export default (state: SharedRecordViewState = initialState, action): SharedReco
         dismissedMatches,
         hiddenMatches
       };
+    case ACTION_TYPES.OPEN_SERVICE:
+      return {
+        ...state,
+        openService: action.payload
```
jkondrat: I would rename this to openedService / selectedService

src/main/webapp/app/modules/conflicts/shared/shared-record-view.reducer.ts
```diff
@@ -75,6 +91,16 @@ export default (state: SharedRecordViewState = initialState, action): SharedReco
...
+      };
+    case ACTION_TYPES.OPEN_LOCATION:
+      return {
+        ...state,
+        openLocation: action.payload
```
jkondrat: I would rename this to openedLocation / selectedLocation (same for method names)

src/main/webapp/app/modules/conflicts/single/details.tsx
```diff
@@ -5,13 +5,15 @@ import { RouteComponentProps, Link } from 'react-router-dom';
 import { IActivityRecord } from 'app/shared/model/activity-record.model';
 import { IOrganizationMatch } from 'app/shared/model//organization-match.model';
 import { OrganizationDetails } from '../shared/components/organization-details';
-import { LocationsDetails } from '../shared/components/location/locations-details';
-import { ServicesDetails } from '../shared/components/service/services-details';
```
jkondrat: what is the point of removing these braces?

src/main/webapp/app/modules/conflicts/single/details.tsx
```diff
@@ -33,6 +35,14 @@ export class Details extends React.Component<ISingleRecordViewProp, ISingleRecor
     isAreaOpen: false
   };
 
+  componentDidMount() {
+    Promise.all([this.props.getNotHiddenMatchesByOrg(this.props.orgId)]).then(() => {
+      if (this.props.matches.length >= 0) {
+        this.props.getPartnerRecords(this.props.orgId);
```
jkondrat: why are we getting all partner records on single record view?

src/main/webapp/i18n/en/service.json
```diff
@@ -42,7 +42,16 @@
       "docs": "Docs",
       "paymentsAccepted": "Payments Accepted",
       "langs": "Langs",
-      "taxonomies": "Taxonomies"
+      "taxonomies": "Taxonomies",
+      "matches": "Matches"
+    },
+    "service_match": {
```
jkondrat: shouldn't we use camelCase?

src/main/java/org/benetech/servicenet/service/mapper/OrganizationMatchMapper.java
```diff
@@ -54,21 +60,27 @@
...
         OrganizationMatchDTO dto = toDto(organizationMatch);
         Organization partner = organizationMatch.getPartnerVersion();
-        Map<UUID, Set<UUID>> locationMatches = new HashMap<>();
+        Map<UUID, List<LocationMatchDto>> locationMatches = new HashMap<>();
+        Map<UUID, List<ServiceMatchDto>> serviceMatches = new HashMap<>();
+
         for (Location location : organizationMatch.getOrganizationRecord().getLocations()) {
-            Set<UUID> matches = new HashSet<>();
-            for (LocationMatch match : locationMatchService.findAllForLocation(location.getId())) {
-                if (partner.getLocations().contains(match.getMatchingLocation())) {
```
jkondrat: Why did we stop filtering these? It was added in #801

src/main/java/org/benetech/servicenet/repository/LocationMatchRepository.java
kdondziak: It's not, but tests don't see without this.

kdondziak: It's not, but tests don't see without this.

src/main/webapp/app/modules/conflicts/single/details.tsx
kdondziak: > what is the point of removing these brackets?

Default import is not connected to redux.

Issue #785, opened 2020-01-13
API: Request and response logging


NeilMBenetech:
As these tasks are being coded as well as all future API-related tasks:
https://github.com/benetech/ServiceNet/issues/782
https://github.com/benetech/ServiceNet/issues/783
https://github.com/benetech/ServiceNet/issues/784

Please log every request and response, including the bearer token used, requesting IP address, the request URL and body, the response HTTP code, URL and Body. This will be useful for us to track usage, potentially enable throttling, and debug any bugs or issues we see over time.

awalkowiak:
Note: The feature is finished, but needs to be moved to ServiceNetGateway service. It will be done when we split db schemas for services #873.

awalkowiak:
@jkondrat Could you confirm that this is working. (excluding the 0x00 issue, which will be resolved in  separate issue)

PR #872, opened 2020-03-10
#785: Add request filter and request logger

awalkowiak:


src/main/java/org/benetech/servicenet/config/FilterConfiguration.java
```diff
@@ -0,0 +1,25 @@
...
+    public RequestFilter requestFilter() {
+        return new RequestFilter();
+    }
+
+    @Bean
+    public FilterRegistrationBean <RequestFilter> filterRegistrationBean() {
+        FilterRegistrationBean <RequestFilter> registrationBean = new FilterRegistrationBean();
+
+        registrationBean.setFilter(requestFilter());
+        registrationBean.addUrlPatterns("/api/*");
```
jkondrat: I wonder if it wouldn't be better to put this filter in gateway service, so that requests from all the other services would also be logged. I think there are also a couple of endpoints having different root url - i believe /management is one of them

src/main/java/org/benetech/servicenet/domain/RequestLogger.java
```diff
@@ -0,0 +1,193 @@
...
+
+    @Lob
+    @Type(type = "org.hibernate.type.TextType")
+    @Column(name = "request_parameters", columnDefinition = "clob")
+    private String requestParameters;
+
+    @Lob
+    @Type(type = "org.hibernate.type.TextType")
+    @Column(name = "request_body", columnDefinition = "clob")
+    private String requestBody;
```
jkondrat: Do we really want to save all request and response bodies? I imagine this could take A LOT of space after some time

src/main/java/org/benetech/servicenet/filters/RequestFilter.java
```diff
@@ -0,0 +1,120 @@
...
+        RequestLogger requestLogger = new RequestLogger();
+        requestLogger.setRemoteAddr(Objects.toString(trace.get("remoteAddr"), ""));
+        requestLogger.setRequestUri(Objects.toString(trace.get("requestUri"), ""));
+        requestLogger.setRequestMethod(Objects.toString(trace.get("requestMethod"), ""));
+        requestLogger.setRequestParameters(Objects.toString(trace.get("requestParams"), ""));
+        requestLogger.setRequestBody(Objects.toString(trace.get("body"), ""));
+        requestLogger.setResponseStatus(Objects.toString(responseTrace.get("status"), ""));
+        requestLogger.setResponseBody(Objects.toString(responseTrace.get("body"), ""));
+
+        requestLoggerService.save(requestLoggerMapper.toDto(requestLogger));
```
jkondrat: I think it's worth measuring if it slows down the requests - maybe we could save them in a separate thread

src/main/java/org/benetech/servicenet/filters/RequestFilter.java
```diff
@@ -0,0 +1,120 @@
...
+            if (buf.length > 0) {
+                String payload;
+                try {
+                    payload = new String(buf, 0, buf.length, wrapper.getCharacterEncoding());
+                }
+                catch (UnsupportedEncodingException ex) {
+                    payload = "[unknown]";
+                }
+
+                trace.put("body", payload.replaceAll("\"password\":\".+?\"", "\"password\":\"fakePassword\""));
```
jkondrat: I wonder if there is more sensitive data; for example we should also filter out the newly changed passwords (I believe this field is called newPassword?)

jkondrat: Is this replace fast enough for big requests? maybe we could only do replacing for some endpoints that we know have this sensitive data

src/main/java/org/benetech/servicenet/filters/RequestFilter.java
```diff
@@ -0,0 +1,120 @@
...
+            if (buf.length > 0) {
+                String payload;
+                try {
+                    payload = new String(buf, 0, buf.length, wrapper.getCharacterEncoding());
+                }
+                catch (UnsupportedEncodingException ex) {
+                    payload = "[unknown]";
+                }
+
+                trace.put("body", payload.replaceAll("\"id_token\" : \".+?\"", "\"id_token\" : \"fakeIdToken\""));
```
jkondrat: do we need to filter out tokens?

src/main/java/org/benetech/servicenet/domain/RequestLogger.java
awalkowiak: Request body is not that bad. The problem might be with response body.

src/main/java/org/benetech/servicenet/filters/RequestFilter.java
awalkowiak: I think it is not as crucial as passwords. I can get rid of it.

awalkowiak:
@jkondrat Ok, ill move this to ServiceNetGateway when the issue related to schema split (#873) is completed.

src/main/java/org/benetech/servicenet/filters/RequestFilter.java
awalkowiak: Replace now will work for authentication and for change-password endpoints.

src/main/java/org/benetech/servicenet/filters/RequestFilter.java
awalkowiak: I do not spot any big difference

src/main/java/org/benetech/servicenet/domain/RequestLogger.java
awalkowiak: Now there is save for response body if the status code is different than 200.

src/main/java/org/benetech/servicenet/config/FilterConfiguration.java
awalkowiak: Done

awalkowiak:
I am closing this PR, since it is moved to Gateway service.

PR #916, opened 2020-04-06
881: Upload data doesn't update organization

kdondziak:


src/main/java/org/benetech/servicenet/domain/Location.java
```diff
@@ -6,27 +6,15 @@
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.List;
-import javax.persistence.JoinColumn;
-import javax.persistence.JoinTable;
-import javax.persistence.ManyToMany;
-import javax.persistence.PrePersist;
-import javax.persistence.PreUpdate;
-import javax.persistence.UniqueConstraint;
+import javax.persistence.*;
```
awalkowiak, line 9: Why '.*'?

src/main/resources/config/application.yml
```diff
@@ -79,6 +79,9 @@ management:
 spring:
   application:
     name: ServiceNet
+  servlet.multipart:
+    max-request-size: 100MB
```
awalkowiak, line 83: 100MB?

PR #924, opened 2020-04-14
314: Timestamp of latest export

kdondziak:


src/main/java/org/benetech/servicenet/service/impl/DataStatusServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Autowired
+    private OrganizationRepository organizationRepository;
+
+    @Override
+    public Page<DataStatusDto> getDataStatuses(Pageable pageable) {
+        List<DataStatusDto> results = new ArrayList<>();
+        Page<SystemAccount> systemAccounts = systemAccountRepository.findAll(pageable);
+        for (SystemAccount systemAccount : systemAccounts) {
+            organizationRepository.findFirstByAccountIdOrderByUpdatedAtDesc(systemAccount.getId())
```
jkondrat, line 35: Shouldn't we actually look at data import reports? I believe that organization's updated at can be modified not only by the imports

Issue #931, opened 2020-04-16
Backend work for allowing users to directly contribute records in ServiceNet


johnhbenetech:
Goal: Allow service providers to contribute information about their own services directly in ServiceNet

The intended approach would NOT involve them “claiming” an existing record, but instead being able to create records from scratch and become “owners” of those records. This is quite similar to the Shelter approach – users create a record, can return to edit it, and also can see other records that have been contributed.

This will interact with our previous work where essentially these user-created records act as a "fake" seventh partner. The matching algorithm should run after each record creation and edit.
These records will appear as matches to the six original partners’ records. We will visually distinguish these records as a sort of “gold standard” record. We will also offer a separate view to all partners of just these records.

This task is to begin required backend work to support this flow, frontend task to follow shortly. As always, let me know if there is a more elegant solution i'm missing. Components affecting the backend (please break out into separate tasks as you see fit):

* Creation of a new system account to which these users will be attached

* By default, newly registered users will be attached to this new system account - they will still be deactivated by default - benetech admins will approve and change system account if necessary

* Extend a similar concept as the "owner" from the Shelter project to these users. When a logged in user creates a record in servicenet, that user should be attached to that record as an "owner" automatically. This will be used to restrict their editing permissions to only the records they created.

* Ensure internal APIs are ready for the creation of a record with the following conditions:
-- all information will likely be added on a single page (can't send them to a location entity page, and then a service entity page, etc). So we may need a new endpoint that deconstructs data into the necessary entities unless it exists already.
-- Allow for the assignment of the ownership via the API
-- Set up the edit API to check for ownership to prevent edits for unauthenticated/non owner users

bognasoldev89:
**Tested with success**

Verified on staging env

**Verified cases**
- Creating a new user account
-- Checking from user management how a new account was created
- Creating a new organization as a new user
-- Checking ownership of a newly created organization

**Summary**
Accounts and organizations are created correctly.

PR #955, opened 2020-04-28
#931: New System Account type for independent service providers

awalkowiak:


src/main/java/org/benetech/servicenet/service/dto/UserDTO.java
```diff
@@ -82,5 +85,10 @@ public UserDTO(UserProfile userProfile) {
                 .map(Shelter::getId)
                 .collect(Collectors.toList());
         }
+        if (userProfile.getOrganizations() != null) {
+            this.shelters = userProfile.getOrganizations().stream()
```
jkondrat, line 89: why are we saving organizations in the shelters field?

awalkowiak: Thanks, Good catch. I missed that.

Issue #783, opened 2020-01-13
API: Request matching records


NeilMBenetech:
Create an an endpoint that only responds to authenticated and authorized API requests (see this related task : https://github.com/benetech/ServiceNet/issues/782 ).

Both the request and the response should be in well-structured JSON. The request should specify the native unique identifier of any Organization, Service or Location record in their own database. 

The response should include an array of matching records of the same record type as the request (O, S, L) with the following attributes:
-Service Net unique identifier (so it can be subsequently requested)
-Native unique identifier (in the owner's database)
-HSDS record type (Organization, Service, Location)
-Record name as present in the partner's datbase (for example "Bob's Food bank")
-Name of the SN partner agency that owns the matching record

In the future we may add "matching" metadata to the response object, but for now we will start with these basic attributes.

kdondziak:
@johnhbenetech I have a question about details of the endpoint. 
1. Should there be one endpoint for (organization, location, services) or one for each of them?
2. Should it accept one id or array of ids?

johnhbenetech:
1. I think we should just start with organizations alone right now. This is the biggest use case by far and I'd rather wait on building anything out for locations or services. 

2. Let's just keep it to a single ID

So essentially there is a get organization matches endpoint

It takes in:
* an organization ID (either a ServiceNet ID or one from their system (external_db_id)
* an optional threshold for similarity (suggestions on what the default should be?) 

It provides back an array of orgs we believe to be matches. each includes
* SN org ID
* SN org Name
* SN org's system account name
* Similarity


Open to suggestions or if you feel im missing anything obvious

kdondziak:
Note for QA:
Test for this issue consists of 2 parts:

1. Creating, reading, updating and deleting client -> see admin menu client management.
2. Getting information from api via postman: 
- Create client in mentioned above client management
- Using credentials get auth token just like #782 
- Using token prepare request:
* Post: https://benetechservicenet.org//services/servicenet/partner-api/records
headers section:
Content-Type: application/json
Cookie: access_token="here paste token"
body section: 
{
    "id": "id of organizatinon from same provider as client e.g. sheltertech 6a5cc3c9-6181-4c6d-ae30-bde10f3a4660",
    "similarity": 0.1
}

abaranowska1:
Following by @kdondziak  instruction I checked:

1. Creating, reading, updating, and deleting external client, and here I found one small issue : 
[...]

PR #962, opened 2020-05-04
 #783: Add unique id error message

kdondziak:


src/main/java/org/benetech/servicenet/service/ClientService.java
```diff
@@ -45,10 +47,14 @@
...
-        ClientDTO result = authClient.createClient(clientDto);
-        this.createOrUpdateClientProfile(result, clientDto);
-        return this.getRealDto(result);
+    public ClientDTO createClient(ClientDTO clientDto) throws BadRequestAlertException {
+        try {
+            ClientDTO result = authClient.createClient(clientDto);
+            this.createOrUpdateClientProfile(result, clientDto);
+            return this.getRealDto(result);
+        } catch (Exception e) {
+            throw new BadRequestAlertException(ErrorConstants.LOGIN_ALREADY_USED_TYPE, "Client id already used!", "clientManagement", "clientexists");
```
jkondrat, line 56: I think we should check for response status, we don't want to throw 'client id already used' for any error
it can be done by creating a custom error decoder, I believe spring picks it up if you make it a @Bean
https://github.com/OpenFeign/feign/wiki/Custom-error-handling

Issue #958, opened 2020-04-29
Integrate Card with carousel


kdondziak:
Integrate Card from this issue: https://github.com/benetech/ServiceNet/issues/944, and carousel from this: https://github.com/benetech/ServiceNet/issues/943
Consider adding separate endpoint and mapper method for this

abaranowska1:
**Tested with success**

Verified on staging env

**Verified cases**

Cards are integrated with the carousel
![image](https://user-images.githubusercontent.com/61964176/81556778-2e9d0100-938b-11ea-935c-b359caea2bdc.png)

PR #977, opened 2020-05-07
 #958: Integrate card with UI

kdondziak:


src/main/java/org/benetech/servicenet/service/dto/ProviderActivityRecordDTO.java
```diff
@@ -0,0 +1,30 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class ProviderActivityRecordDTO {
```
jkondrat, line 14: let's rename it to ProviderRecordDTO

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -151,4 +172,13 @@ private ActivityDTO getEntityActivity(ActivityInfo info, Map<UUID, ExclusionsCon
             return new PageImpl<>(Collections.emptyList(), pageable, Collections.emptyList().size());
         }
     }
+
+    private ProviderActivityRecordDTO mapToProviderActivityRecordDTO(Organization organization) {
+        try {
+            Optional<ActivityRecordDTO> opt = recordsService.getRecordFromOrganization(organization);
```
jkondrat, line 178: is it possible to only get necessary data from the database instead of mapping activity records? or would it take too much effort?

Issue #956, opened 2020-04-28
Move unique together constraint for organization check from db to import service


awalkowiak:
Currently, there is unique  together (external_db_id, account_id) for organization. This should not be checked for new users with the "Service Provider" system account. Remove this constraint from the database and move it to the import service. This should be checked only during the file upload or getting data through external API

(related issue: #931)

bognasoldev89:
**Tested with success**

Verified on staging env

**Verified cases**
- adding organization via Service Provider user
- checking this organization on Organization entity
- comparing this to "classic" organization record

**Attachment**
- Organization entity view
![sn 956 1](https://user-images.githubusercontent.com/51324797/83014614-547b0480-a01f-11ea-8924-fa82be43f5d6.gif)

- Organization added via Service Provider user
![sn 956 2](https://user-images.githubusercontent.com/51324797/83014708-78d6e100-a01f-11ea-9fc2-0f0aeea885aa.png)

PR #1020, opened 2020-05-22
#956: Remove unique together constraint on organization's external db…

jkondrat:
… id and account name


awalkowiak:
Lgtm, but checks failed

awalkowiak:
@jkondrat conflicts

Issue #1033, opened 2020-05-28
Allow for creation of separate silos of the ServiceProvider view (for different regions as an example).


kkaczmarczyk:
* A user entity should be associated with a specific silo
* Admin users should be able to create new named silos, that can later be used to assign users

dev note: consider using jhipster auto-generated views

abaranowska1:
@kdondziak 
Please verify and fix these cases:

1. Missing translations in Silo Entity: 
Create, Update, Delete 

![image](https://user-images.githubusercontent.com/61964176/84007906-58891980-a971-11ea-9356-2132d40f17d6.png)
![image](https://user-images.githubusercontent.com/61964176/84007916-5b840a00-a971-11ea-9e87-ac2a6fc52852.png)e
![image](https://user-images.githubusercontent.com/61964176/84007928-5fb02780-a971-11ea-9659-777e99090a0c.png)

2. After removing the silo we are moving to the HomePage, but in the other Entity's we stay on the opened Entity page.
![image](https://user-images.githubusercontent.com/61964176/84008345-0a284a80-a972-11ea-804a-5270ef5bb6cf.png)

I checked also assigning users to the Silo and here from my side all works fine. For example user: **provider** and user: **bognaprov** are in the same silo. Both users have access to see records only from the same silo.
![image](https://user-images.githubusercontent.com/61964176/84009179-31334c00-a973-11ea-901d-f997babe0c60.png)
![image](https://user-images.githubusercontent.com/61964176/84009201-37c1c380-a973-11ea-9f30-42593ad4910c.png)

abaranowska1:
**Tested with success**

- Missing translations in Silo Entity: Create, Update, Delete were fixed
- After removing the silo we are staying on the opened Entity page.

PR #1058, opened 2020-06-09
 #1033: Add silo name unique check

kdondziak:


src/main/java/org/benetech/servicenet/web/rest/SiloResource.java
```diff
@@ -87,6 +92,13 @@ public SiloResource(SiloService siloService) {
         if (siloDTO.getId() == null) {
             throw new BadRequestAlertException("Invalid id", ENTITY_NAME, "idnull");
         }
+
+        siloService.findOneByName(siloDTO.getName()).ifPresent(s -> {
+            if (!s.getId().equals(siloDTO.getId())) {
+                throw new BadRequestAlertException("Name is already used", ENTITY_NAME, "nameunique");
```
jkondrat, line 98: why do we need a second check?

src/main/java/org/benetech/servicenet/web/rest/SiloResource.java
```diff
@@ -65,6 +65,11 @@ public SiloResource(SiloService siloService) {
         if (siloDTO.getId() != null) {
             throw new BadRequestAlertException("A new silo cannot already have an ID", ENTITY_NAME, "idexists");
         }
+
+        siloService.findOneByName(siloDTO.getName()).ifPresent(s -> {
```
jkondrat, line 69: perhaps we could also add a database constraint

Issue #1081, opened 2020-07-02
New user profile fields should be available in user settings page as well for a logged in user


johnhbenetech:
In #1071 we added several new fields. These fields should also be available for logged in users on the settings page.

One thing to call out is that we are now requiring org name to be filled out, but obviously there exist users who were registered prior to that change. We should expect that on load that field may be empty, but should enforce the requirement on any subsequent change

abaranowska1:
**Tested with success**
- New user profile fields are available on user settings page as well.
- User can edit data from the settings page and all changes are also available for admin 
- Changes made by admin are visible for user 
- Data from the registration are shown on user settings page 
![Peek 2020-07-03 09-51](https://user-images.githubusercontent.com/61964176/86446130-cbd53f80-bd13-11ea-89e6-140e007633fe.gif)

PR #1082, opened 2020-07-02
#1081: Add new UserProfile fields on user settings page

awalkowiak:


src/main/java/org/benetech/servicenet/web/rest/UserResource.java
```diff
@@ -214,4 +214,22 @@ public UserDTO getAccount() {
...
             .body(user);
     }
+
+    /**
+     * {@code POST  /account : update user account.
+     *
+     * @return the updated current user.
+     * @throws RuntimeException {@code 500 (Internal Server Error)} if the user couldn't be returned.
+     */
+    @PostMapping("/account")
```
kdondziak, line 224: If this is updating can you use PUT method instead of POST?

Issue #1054, opened 2020-06-04
Add 'public' option to silo and allow admin to set it


kdondziak:

bognasoldev89:
**Tested with success**

**Environment**
Staging env

**Verified cases**
- Creating public silo
- Creating not public silo
- Changing silo from public to not public
- Changing silo from not public to public

**Summary**
A public option for silo is available in the Silos entity.

**Attachment**
![sn 1054](https://user-images.githubusercontent.com/51324797/87047249-896f9f00-c1fa-11ea-9163-7e0ff2a77260.gif)

PR #1090, opened 2020-07-09
benetech/ServiceNet#1054: Add public field to silo

kdondziak:


src/main/java/org/benetech/servicenet/domain/Silo.java
```diff
@@ -75,6 +80,14 @@ public Silo userProfiles(Set<UserProfile> userProfiles) {
         return this;
     }
 
+    public boolean isPublic() {
+        return isPublic;
+    }
+
+    public void setPublic(boolean aPublic) {
```
awalkowiak, line 87: Why `aPublic`?

src/main/java/org/benetech/servicenet/domain/Silo.java
```diff
@@ -75,6 +80,14 @@ public Silo userProfiles(Set<UserProfile> userProfiles) {
...
+        isPublic = aPublic;
```
awalkowiak, line 88: Why without `this`?

Issue #1074, opened 2020-07-01
Improve records loading on map view


awalkowiak:
Currently, the map view is getting records from the same endpoint as the grid view. It should have a separate endpoint with separate DTO (with fields it only requires). Also, map view and grid view records should have separated reducers state (when going back from the map to the grid view we should get the same records amount - page as before going to the map view).

bognasoldev89:
**Tested with success**

**Environment**
- Staging env
- Browsers
-- Chrome 83.0.4103.116

**Verified cases**
- Changing view from grid to map and back with different amount of records loaded
- Comparing console for organizations on map and grid view

**Summary**
The map view shows pins for all of the available records, but the amount of loaded records doesn't change. The console view confirms the different DTO.

**Attachments**
- Changing views between grid and map
![sn 1074 map](https://user-images.githubusercontent.com/51324797/87768736-e1bb2800-c81c-11ea-8cfd-257801bb6092.gif)

- Console for grid view
![sn 1074 grid view](https://user-images.githubusercontent.com/51324797/87768754-e97acc80-c81c-11ea-9bb7-42ca321a3846.png)

- Console for map view
![sn 1074 console map view](https://user-images.githubusercontent.com/51324797/87768763-eda6ea00-c81c-11ea-9ab0-633acfad54b2.png)

PR #1091, opened 2020-07-10
#1074: Improve records loading on map view

kdondziak:


src/main/java/org/benetech/servicenet/service/ActivityService.java
```diff
@@ -31,5 +32,9 @@
 
     Page<ProviderRecordDTO> getAllPartnerActivities(ProviderFilterDTO providerFilterDTO, String search, Pageable pageable);
 
+    Page<ProviderRecordForMapDTO> getAllPartnerActivitiesForMap(ProviderFilterDTO providerFilterDTO, String search, Pageable pageable);
```
awalkowiak, line 35: Do we need pagination in this endpoint?

Issue #1045, opened 2020-06-02
Create a setting that allows certain silos to expose a public facing list of records


johnhbenetech:
In the silo admin page, create a setting that allows for a public (unauthenticated) view of the records. There will be silos with this setting off, and off should be the default.

This view will be very similar to the current page, but just will not have the 'Your record' section at the top (therefore no ability to add records). Once again this is a 'read-only' view.

If possible this should automatically create the endpoint for this as well. Let me know what's possible here, for instance:

chicago.servicenet.benetech.org
or
servicenet.benetech.org/#/chicago
or
whatever else might be best approach

kdondziak:
Creating public option will be handled in https://github.com/benetech/ServiceNet/issues/1054

kdondziak:
@johnhbenetech Should unauthorized user have access to single record view (when clicking on link on record card)?

johnhbenetech:
Yes


________________________________

kdondziak:
How to test:
1. Set some silo as public.
2. Log out and got to /public/"Name of the silo"

abaranowska1:
@kdondziak Please take a look for a small UI issue: 
- Mobile: Logo is not centered and I think we can move the whole section up or do the same bar as on web view 
![image](https://user-images.githubusercontent.com/61964176/88279908-20515680-cce5-11ea-85a5-50d8a144e96c.png)
![image](https://user-images.githubusercontent.com/61964176/88280061-758d6800-cce5-11ea-8f16-fb1a87b22855.png)
- Firefox issue: I don't know why we have doubled scroll bar, could you check it? 
![Peek 2020-07-23 12-49](https://user-images.githubusercontent.com/61964176/88280268-c8671f80-cce5-11ea-8731-20f9f8475221.gif)

@johnhbenetech URL address is sensitive to small and capital letter in the name of Silo, is it ok for you?

johnhbenetech:
Case-sensitive url is fine

abaranowska1:
@kdondziak 
Please take a look at the mobile view. If you can, center benetech logo and I think we can move the whole section up or do the same bar as on web view.
![Selection_534](https://user-images.githubusercontent.com/61964176/88377425-6f5bc200-cd9f-11ea-98f4-65fac632afa7.png)

Firefox issue was fixed. There is no doubled scroll bar. 
![image](https://user-images.githubusercontent.com/61964176/88377386-5c48f200-cd9f-11ea-9f6c-d45ded68d231.png)

[...]

PR #1097, opened 2020-07-21
#1045: Create a setting that allows certain silos to expose a public …

kdondziak:
…facing list of records

### DIFF
--- a/src/main/java/org/benetech/servicenet/repository/ProviderRecordsRepository.java
+++ b/src/main/java/org/benetech/servicenet/repository/ProviderRecordsRepository.java
@@ -86,7 +86,39 @@ public class ProviderRecordsRepository {
         }
 
         List<Organization> results = query.getResultList();
-        Long total = this.getTotal(providerFilterDTO, userProfile, search);
+
+        Long total = 0L;
+        if (results != null && results.size() > 0) {
+            total = this.getTotal(providerFilterDTO, userProfile, search);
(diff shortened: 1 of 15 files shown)


src/main/java/org/benetech/servicenet/repository/ProviderRecordsRepository.java
```diff
@@ -225,6 +312,14 @@ private Long getTotal(ProviderFilterDTO providerFilterDTO, UserProfile userProfi
         return em.createQuery(countCriteria).getSingleResult();
     }
 
+    private Long getTotal(ProviderFilterDTO providerFilterDTO, Silo silo, String search) {
+        CriteriaQuery<Long> countCriteria = cb.createQuery(Long.class);
+        Root<Organization> selectRoot = countCriteria.from(Organization.class);
+        this.addFilters(countCriteria, selectRoot, silo, providerFilterDTO, search);
+        countCriteria.select(cb.countDistinct(selectRoot));
+        return em.createQuery(countCriteria).getSingleResult();
```
awalkowiak, line 320: I think that when em.createQuery(countCriteria) will be empty, then .getSingleResult() will throw nullpointer. Same case as it happened here: https://github.com/benetech/ServiceNet/blob/dev/src/main/java/org/benetech/servicenet/repository/ShelterRepositoryImpl.java#L101

Issue #1108, opened 2020-07-25
Map view on ServiceProvider view should respect filters


johnhbenetech:
Pins should only be shown from organizations that have at least one location/service that satisfies the selected filter

abaranowska1:
@kdondziak 
I found some issues:
1. When Organization have more than one location, but we filtered only by City: for example, Chicago - than only Chicago pins should be displayed
![image](https://user-images.githubusercontent.com/61964176/89191541-5fb05a80-d5a3-11ea-95b7-2d0bf55b76d9.png)

2. After clicking Clear button ( on the Map View ) I can see that only the map is cleared, but records stay with filters. After come back to the Grid View I can still see the filtered data. 
![Peek 2020-08-03 15-48](https://user-images.githubusercontent.com/61964176/89191780-b74ec600-d5a3-11ea-9568-14a50a5a1bb2.gif)
![image](https://user-images.githubusercontent.com/61964176/89191799-bd44a700-d5a3-11ea-9a4b-001c5e2e2527.png)

3. Similar issue, after use filters on Map View the only map is changing but records number stays 
![image](https://user-images.githubusercontent.com/61964176/89192390-9a66c280-d5a4-11ea-8001-e8db33ab45c0.png)

kdondziak:
@johnhbenetech Let's say we have a use case:
1. Change filters on the grid view.
2. Switch to map view and change filters (different than in point 1)
3. Switch back to grid view
Correct me if I am wrong, but it was supposed to show same records on the grid view regardless of filters change on the map view? Should those 2 filters work independently?

johnhbenetech:
@kdondziak the filters should affect both views in the same way. Does that make sense?

abaranowska1:
@kdondziak 
- On provider account, I picked City: Los Angels filter and I noticed an issue with Brand New Org , ( pin near Teksas shouldn't be displayed) , plz check it 
![image](https://user-images.githubusercontent.com/61964176/89640309-98a44400-d8af-11ea-8ef6-c5eaf8fcecef.png)
![image](https://user-images.githubusercontent.com/61964176/89640356-b07bc800-d8af-11ea-8790-991ff9525f17.png)

- According to #1027 I remember that on the filters we only have what is on the list of other providers from my silo. Also only real data will be displayed. I picked City: Oakland or Louisville and 0 records was found but on the map view the Ocean is displayed 
[...]

PR #1138, opened 2020-08-25
#1108: Refactor filter cities repository

kdondziak:


src/main/java/org/benetech/servicenet/repository/ActivityFilterCityRepository.java
```diff
@@ -0,0 +1,95 @@
...
+
+        addFiltersAndSelectForCities(queryCriteria, selectRoot, userProfile, silo);
+
+        Query query = em.createQuery(queryCriteria);
+
+        List<String> results = query.getResultList();
+        Set<String> result = new HashSet<>();
+        if (results != null) {
+            for (String locality : results) {
+                if (StringUtils.isNotBlank(locality)) {
```
pmuchowski, line 62: I think you can add this to the query

pmuchowski (changes requested):

pmuchowski (approved):

Issue #1161, opened 2020-08-31
Add fake/reference data for SP view


kdondziak:
Devs working on SP view have too small a data set. The goal of this issue is to add sql script/migration to add enough amount of data to experience possible performance issues.

kdondziak:
@jkondrat @awalkowiak @pmuchowski usage: 
If you want to add organizations to specific user
./mvnw -Pgen  -Dspring-boot.run.arguments="--login=userLogin --orgs=5"
if you want to add organizations to all users in the system (n orgs to each user) 
./mvnw -Pgen  -Dspring-boot.run.arguments="--orgs=5"

PR #1201, opened 2020-09-10
benetech/ServiceNet#1161: Add fake data generator

kdondziak:


src/main/java/org/benetech/servicenet/scheduler/CreateReferenceDataJob.java
```diff
@@ -0,0 +1,48 @@
...
+import org.benetech.servicenet.repository.UserProfileRepository;
+import org.benetech.servicenet.service.SystemAccountService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.transaction.annotation.Transactional;
+
+@org.springframework.stereotype.Service
+@Transactional
+public class CreateReferenceDataJob {
```
jkondrat, line 19: let's call it ReferenceDataGenerator

jkondrat (approved):

src/main/java/org/benetech/servicenet/ServiceNetApp.java
```diff
@@ -65,6 +71,14 @@ public void initApplication() {
             LOG.error("You have misconfigured your application! It should not " +
                 "run with both the 'dev' and 'cloud' profiles at the same time.");
         }
+
+        if (env.getProperty("login") != null && env.getProperty("orgs") != null) {
+            String userLogin = env.getProperty("login");
```
jkondrat, line 76: if we don't provide user's login, let's distribute the organizations across random users

jkondrat (changes requested):

jkondrat (approved):

PR #1250, opened 2020-09-21
Use DTO when fetching Provider records

jkondrat:


src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -151,33 +151,32 @@ public ActivityServiceImpl(ActivityRepository activityRepository, RecordsService
...
         if (userGroups == null || userGroups.isEmpty()) {
-            organizations = providerRecordsRepository
+            return providerRecordsRepository
                 .findAllWithFilters(userProfile, null, new ProviderFilterDTO(), null, pageable);
         } else {
-            organizations = organizationService.findAllByUserGroups(pageable, new ArrayList<>(userGroups));
+            return organizationService.findAllByUserGroups(pageable, new ArrayList<>(userGroups))
+                .map(this::getProviderRecordDTO);
         }
-        return organizations.map(this::getProviderRecordDTO);
```
pmuchowski, line 161: filterProviderRecords should be used here

awalkowiak (approved):

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -151,33 +151,34 @@ public ActivityServiceImpl(ActivityRepository activityRepository, RecordsService
     public Page<ProviderRecordDTO> getPartnerActivitiesForCurrentUser(Pageable pageable) {
         UserProfile userProfile = userService.getCurrentUserProfile();
         Set<UserGroup> userGroups = userProfile.getUserGroups();
-        Page<Organization> organizations;
         if (userGroups == null || userGroups.isEmpty()) {
-            organizations = providerRecordsRepository
+            return providerRecordsRepository
```
pmuchowski, line 155: you should use the filterProviderRecords here (to remove the excluded fields and locations)

src/main/java/org/benetech/servicenet/service/impl/ActivityServiceImpl.java
```diff
@@ -151,33 +151,34 @@ public ActivityServiceImpl(ActivityRepository activityRepository, RecordsService
...
                 .findAllWithFilters(userProfile, null, new ProviderFilterDTO(), null, pageable);
         } else {
-            organizations = organizationService.findAllByUserGroups(pageable, new ArrayList<>(userGroups));
+            return filterProviderRecords(
+                organizationService.findAllByUserGroups(pageable, new ArrayList<>(userGroups))
+                .map(this::getProviderRecordDTO)
```
pmuchowski, line 160: filter is not needed here, because getProviderRecordDTO is removing the excluded fields

Issue #1260, opened 2020-09-25
Reinstating records after dismissing cause an internal server error


abaranowska1:
During regression testing, I met an issue with reinstating records. Internal Server Error is thrown after reinstating records a few times. 
![Peek 2020-09-25 09-14](https://user-images.githubusercontent.com/61964176/94248919-729a3800-ff1f-11ea-821c-056333be7d19.gif)
![Screenshot_2020-09-25_09-10-50](https://user-images.githubusercontent.com/61964176/94248988-89408f00-ff1f-11ea-8894-cd7b6fcca4fa.png)
![Screenshot_2020-09-25_09-11-03](https://user-images.githubusercontent.com/61964176/94248998-8a71bc00-ff1f-11ea-922d-aa862c2fd1ca.png)
![Screenshot_2020-09-25_09-11-43](https://user-images.githubusercontent.com/61964176/94249001-8b0a5280-ff1f-11ea-8956-ac808a53d5e9.png)
![Screenshot_2020-09-25_09-11-58](https://user-images.githubusercontent.com/61964176/94249003-8ba2e900-ff1f-11ea-9f64-c1b0ef74de5f.png)

dev note: 
`conflictDetectionService.detectAsynchronously(match.getOrganizationRecord(), Collections.singletonList(match));`

abaranowska1:
**Tested with success**

It is possible to reinstate records after dismissing without issues.

![Peek 2020-09-29 10-05](https://user-images.githubusercontent.com/61964176/94530556-8acdc800-023b-11eb-921b-0032ca80fa60.gif)

**Verified on:**

- Ubuntu 20.04
- Google Chrome v. 85.0.4183.83

PR #1261, opened 2020-09-25
benetech/ServiceNet#1260: Reinstate removed conflicts instead of dete…

jkondrat:
…cting them again


src/main/java/org/benetech/servicenet/service/impl/OrganizationMatchServiceImpl.java
```diff
@@ -318,7 +317,7 @@ public void revertDismissOrganizationMatch(UUID id) {
 
             organizationMatchRepository.save(match);
 
-            conflictDetectionService.detectAsynchronously(match.getOrganizationRecord(), Collections.singletonList(match));
+            conflictDetectionService.reinstate(match);
```
awalkowiak, line 320: This potentially could be a good change, but I see here an edge case: when you reinstate a conflict between two records after one of them being completely changed and the conflict is obsolete and gives false/old info.

Issue #1149, opened 2020-08-27
Adding Hours of operation to ServiceProvider


johnhbenetech:
- add the below "hours of operations" table in editing a record under location details 
- ideally, every location they have will have each own hours of operations time list
- when adding locations, a new table should show up for each location a user adds

<img width="320" alt="Screenshot 2020-11-30 154610" src="https://user-images.githubusercontent.com/65787629/100679522-88174e00-3324-11eb-9771-f9130fb7ce85.png">

**service Net theme**
<img width="350" alt="Screenshot 2020-11-30 155604" src="https://user-images.githubusercontent.com/65787629/100679560-97969700-3324-11eb-92a7-3391a0c39f89.png">

Changes on update my record (desktop)
---------------------------
- add below the address

<img width="662" alt="Screenshot 2020-11-30 160906" src="https://user-images.githubusercontent.com/65787629/100680443-661ecb00-3326-11eb-9546-b0c9e93a8ddc.png">

Changes on Tell us where your services are Located in adding a new record (desktop)
--------------------------------
<img width="752" alt="Screenshot 2020-11-30 155352" src="https://user-images.githubusercontent.com/65787629/100679436-5900dc80-3324-11eb-9258-f234e27589cb.png">

Changes on update my record (mobile)
---------------------------
<img width="140" alt="Screenshot 2020-11-30 160515" src="https://user-images.githubusercontent.com/65787629/100680181-db3dd080-3325-11eb-92a1-2f47bcd30b28.png">

Changes on Tell us where your services are Located in adding a new record (mobile)
--------------------------------
<img width="143" alt="Screenshot 2020-11-30 160456" src="https://user-images.githubusercontent.com/65787629/100680153-d1b46880-3325-11eb-9133-71bcf8f9ab97.png">

**Database**
- add hours of operation in the database

EnriqueCorado:
@meenupant

johnhbenetech:
@meenupant @EnriqueCorado What's in the time dropdowns?

Are we sure this should be at the location level and not the service level? What is that assumption based on?

Often times you will see indications of something like a lunch break - e.g. open from 9-12, closed 12-2, open 2-5 - do we want to deal with this case?

EnriqueCorado:
@johnhbenetech under the time dropdowns
[...]

PR #1469, opened 2020-12-14
benetech/ServiceNet#1149: Adding Hours of operation to ServiceProvider

jkondrat:


src/main/java/org/benetech/servicenet/service/dto/provider/ProviderOpeningHoursDTO.java
```diff
@@ -0,0 +1,86 @@
...
+    private UUID id;
+
+    @NotNull
+    private Integer weekday;
+
+    private String opensAt;
+
+    private String closesAt;
+
+    public UUID getId() {
```
awalkowiak, line 24: Here you use getters/setters and in another case Lombok's `@Data` and in yet another case Lombok's `@Getters/@Setters`. Perhaps we should be more consistent with this.

awalkowiak (approved):
LGTM. There is one thing, that I noticed but this can be moved to the separate refactor/cleanup issue.

Issue #1446, opened 2020-12-02
Claim a record 


EnriqueCorado:
The flow will be
- when a user register we will shortlist organizations close to the name that they registered with during the registration 
- once the user gets manually approved by Benetech staff
- User during their first log in will have a pop up that shows records that match the information provided in the registration 
- users can have the ability to claim the ones they will manage 

**New users will initially see this display for a few seconds**
<img width="959" alt="Screenshot 2020-12-02 153713" src="https://user-images.githubusercontent.com/65787629/100945149-8335d400-34b5-11eb-9902-1004e60edf54.png">

**After a few seconds a pop up **
<img width="890" alt="Screenshot 2020-12-02 154954" src="https://user-images.githubusercontent.com/65787629/100945423-0ce5a180-34b6-11eb-9cdf-eb21704a88ed.png">

<img width="892" alt="Screenshot 2020-12-02 155016" src="https://user-images.githubusercontent.com/65787629/100945443-15d67300-34b6-11eb-90e8-f03c56959036.png">

- after users have claimed records, or didn't claim any of the records, they can press done and the records claimed will show up on their records

EnriqueCorado:
@meenupant

meenupant:
@EnriqueCorado What if the records that they should have claimed but do not pop up for some reason (maybe a slight spelling mistake or incorrect location)? We would want them to have another option saying "Did we miss finding your record(s)? Would you like to search your record(s) by either location or organization name?" And another option "Would you like to claim more record(s)?" And once they claim their record(s) then we would want them to review all the data fields one by one and give them an option to either confirm (the correct entries) or modify (the incorrect ones). 
And once they are done with this process, let them preview all the information before final confirmation.
@johnhbenetech Did we miss anything?

EnriqueCorado:
- on users first, log in, they will have the below pop up 
- the records will be pre-populated from the organization name(s) matching records on the service net

<img width="744" alt="Screenshot 2020-12-03 121534" src="https://user-images.githubusercontent.com/65787629/101083040-41626780-3561-11eb-8c21-daaef80454a5.png">


- users will be able to claim those records whom are theirs 
- once the user clicks "Done" those records will appear as any record those for a service provider
<img width="751" alt="Screenshot 2020-12-03 121644" src="https://user-images.githubusercontent.com/65787629/101083218-6bb42500-3561-11eb-8cbc-ac102fc905f9.png">


[...]

Issue #1333, opened 2020-10-15
Unclaiming a record 


EnriqueCorado:
This will throw out any user-provided info and just re-add the original Eden record to the collection
- make sure we add this to the schema in case we don't add this feature immediately, but have the functionality ready to go 
- add button at the bottom, same font, size, color 

<img width="729" alt="Screenshot 2020-10-15 154756" src="https://user-images.githubusercontent.com/65787629/96193785-d18b1580-0efd-11eb-8824-0eaa00820816.png">

meenupant:
How is Unclaim Record different than Deactivate Record? I want to understand the process when someone clicks one of these buttons then what happens to the record. The user may get confused and this may delay any action taken by him. The CTA must be absolutely clear to the user.

EnriqueCorado:
@meenupant unclaim breaks ties from the user updating the record and would go back to the algorithm updating the record 
Deactivate record is for users who created their own record which did not exist in service net so they can essentially delete it

meenupant:
@EnriqueCorado Do they really care if it exists in our records? It is up to us to decide whether we should just revert back to the old record that already exists in our database or just remove it if it does not exist. I think let's brainstorm this further to make the process more clear and not confuse the user.

kkaczmarczyk:
@EnriqueCorado @meenupant should some pop-up appear after clicking on "Unclaim Record" button? Also, are there any new findings coming from the brainstorm?

EnriqueCorado:
@kkaczmarczyk 
- have a pop up saying "Record successfully unclaimed" 
- no new findings from brainstorm

abaranowska1:
@awalkowiak 
The unclaim button disappear after I did any update and click the I'm Finished button 
![Peek 2020-12-16 14-10](https://user-images.githubusercontent.com/61964176/102352824-8ac2a600-3fa8-11eb-93ef-bcd03023ea4c.gif)

abaranowska1:
**Tested with success**
- The unclaim button stays after update record 
![image](https://user-images.githubusercontent.com/61964176/102371304-5659e480-3fbe-11eb-86fb-f1ec741f59ad.png)

PR #1480, opened 2020-12-16
#1446, #1333: Add option to claim and unclaim records

awalkowiak:


src/main/java/org/benetech/servicenet/domain/UserProfile.java
```diff
@@ -109,6 +109,8 @@
     @Column(name = "avatar_base64", columnDefinition = "clob")
     private String avatarBase64;
 
+    private Boolean claimedRecords = false;
```
jkondrat, line 112: rename to hasClaimedRecords

src/main/java/org/benetech/servicenet/service/impl/OrganizationServiceImpl.java
```diff
@@ -547,6 +549,65 @@ public void reactivate(UUID id) {
...
+        userService.saveProfile(currentUserProfile);
+    }
+
+    @Override
+    public void unclaimRecord(UUID recordToUnclaim) {
+        Organization orgToUnclaim = organizationRepository.getOne(recordToUnclaim);
+        Organization originalOrg = organizationRepository.getOne(orgToUnclaim.getReplacedBy().getId());
+        originalOrg.setReplacedBy(null);
+
+        UserProfile userProfile = userService.getCurrentUserProfile();
```
jkondrat, line 604: when the user has no records claimed, shouldn't we set the boolean in userprofile to false? maybe it would be better to have a relation with these organizations and have a boolean (claimedRecords.size() == 0) in dto

jkondrat (changes requested):

src/main/java/org/benetech/servicenet/service/impl/OrganizationServiceImpl.java
awalkowiak: I think for now there is no need for it. The existing `hasClaimedRecords` boolean indicates if the user has ever done the claiming (this is mainly for the frontend), no other checks needed at this point.

Issue #1537, opened 2021-01-04
New SystemAccount for LinkForCare: allow for import via job UI


EnriqueCorado:
Description:
Add a new SystemAccount for LinkForCare and allow for import via job UI

- in the excel file shared there are two tabs: Data & Mapping
- the mapping tab of the excel sheet shows the CSV current column name and what it maps to in SN
- there is also a table to the right of it that has a summary of the SN fields as well as new fields with their task number which are in the up next column. Those are:

COVID 19   Protocols (task #1535)
Service Fee   (task #1449)
replace the "Last update" section in SN with the field LASTMODIFIEDDATE from the CSV file
replace  "By" section of the record with LinkforCare

<img width="307" alt="Screenshot 2021-01-04 150915" src="https://user-images.githubusercontent.com/65787629/103588830-f197f880-4e9e-11eb-9b0d-de816660d03f.png">

Phone Number   (task #1442)
24 / 7 Opening   checkbox in Opening Hours (task #1536)
Remote Service   Checkbox (task #1321)

Acceptance Criteria:
*

Dev Signoffs (remove items if not applicable to this issue)
- [ ] Unit test added for the core business features at the minimum
- [ ] Code review passed
- [ ] CI passed successfully
- [ ] CodeQL scan for vulnerabilities in code has passed 
- [ ] Docker images have been built and deployed to the dev server

QA Signoffs (remove items if not applicable to this issue)
- [x] Passes Acceptance Criteria
- [x] All bug reports fixed/merged/tested
- [x] Regression test cases cycle updated if needed

Product Signoff (remove items if not applicable to this issue)
- [x] Approved for release

EnriqueCorado:
@johnhbenetech

johnhbenetech:
Metadata notes:

We can use the ID, LastModified, and Createdby columns from the CSV for those corresponding fields in ServiceNet

jkondrat:
@kdondziak  please import the online/remote service option as a part of #1321

abaranowska1:
**Tested**
The following tasks are done and the import via job UI is possible. The whole mapping rules were added to the SN and now we can upload data via UI for new LinkForCare provider. For small bugs found during testing this ticket, we were created new tasks. 
- COVID 19 Protocols (task #1535)
- Service Fee (task #1449)
- Phone Number (task #1442)
- 24 / 7 Opening checkbox in Opening Hours (task #1536)

PR #1612, opened 2021-01-15
benetech/ServiceNet#1537: Handle xlsx TRUE values; remove the leading…

jkondrat:
… semicolon


jkondrat:
it turns out that Excel and LibreOffice handle TRUE values differently; handle both

kdondziak (approved):

Issue #1610, opened 2021-01-15
When claiming a record, include the new fields used by LinkForCare import


jkondrat:
Make sure that all the fields are copied when claiming a record

jkondrat:
@abaranowska1 It seems that all the (existing) fields are already included, could you please confirm?

@kdondziak when adding remote service checkbox (#1321) please remember to verify that it works with claiming

abaranowska1:
@jkondrat it seems that not every fields are included after claim. 
PRIMARY_CONTACT_PHONE__C, "Which Locations Offer This Service?" (location in the service is not selected) 
![image](https://user-images.githubusercontent.com/61964176/104898575-19e12600-597a-11eb-8752-4cf86cc5bb59.png)
![image](https://user-images.githubusercontent.com/61964176/104898609-25345180-597a-11eb-86fb-f5f72b9c8132.png)

abaranowska1:
**Tested with success**
"Which Locations Offer This Service?"  and "Phone number" fields are copied after claim. 
![image](https://user-images.githubusercontent.com/61964176/105186362-ecd77380-5b31-11eb-8550-d38bed377637.png)
Note: I should update the record via import firstly.

PR #1639, opened 2021-01-19
benetech/ServiceNet#1610: Include ServiceAtLocation when claiming an …

jkondrat:
…organization

### DIFF
--- a/src/main/java/org/benetech/servicenet/service/impl/OrganizationServiceImpl.java
+++ b/src/main/java/org/benetech/servicenet/service/impl/OrganizationServiceImpl.java
@@ -612,9 +612,9 @@ public class OrganizationServiceImpl implements OrganizationService {
         orgClone.setReplacedBy(null);
         organizationRepository.save(orgClone);
 
-        Set<Service> clonedServices = cloneServices(services, orgClone);
-
         Set<Location> clonedLocations = cloneLocations(locations, orgClone);
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/benetech/servicenet/service/impl/OrganizationServiceImpl.java
```diff
@@ -1112,7 +1112,24 @@ private void saveDatesClosed(Map<Integer, List<String>> datesClosedByLocation, L
...
+            Set<ServiceAtLocation> sats = new HashSet<>();
+            for (ServiceAtLocation sat : service.getLocations()) {
+                Optional<Location> loc = locations.stream()
+                    .filter(l -> l.getExternalDbId() != null
+                        && l.getExternalDbId().startsWith(sat.getLocation().getExternalDbId()))
+                    .findFirst();
+                if (loc.isPresent()) {
+                    ServiceAtLocation serviceAtLocation = new ServiceAtLocation();
+                    serviceAtLocation.setProviderName(SERVICE_PROVIDER);
+                    serviceAtLocation.setLocation(loc.get());
```
awalkowiak, line 1126: I think the cloned location should be set here instead of the original service's location.

jkondrat: the cloned locations are passed as a 'locations' param

awalkowiak (approved):

